PR #2, opened 2020-05-18
Ra dspdc 1022 hca process metadata

raaidbroad:
- Add build.sbt and create project structure
- Move schema from mvp.json to individual {entity_type}.table.json files
- Implement pipeline pieces for non-file metadata entities


transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
```diff
@@ -0,0 +1,36 @@
+package org.broadinstitute.monster.hca
+
+import com.spotify.scio.ScioContext
+import org.broadinstitute.monster.common.PipelineBuilder
+
+object HcaPipelineBuilder extends PipelineBuilder[Args] {
+
+  override def buildPipeline(ctx: ScioContext, args: Args): Unit = {
+    // process metadata for each non-file metadata entity
+    Set(
```
danxmoran, line 10: Would make this `Set` a static `val` at the object level

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaUtils.scala
```diff
@@ -0,0 +1,119 @@
...
+import org.apache.beam.sdk.io.{FileIO, ReadableFileCoder}
+import org.apache.beam.sdk.io.fs.EmptyMatchTreatment
+import org.broadinstitute.monster.common.StorageIO
+import org.broadinstitute.monster.common.msg.{JsonParser, UpackMsgCoder}
+import upack.{Msg, Obj, Str}
+
+import scala.collection.mutable
+import scala.util.matching.Regex
+
+object HcaUtils {
```
danxmoran, line 17: I don't see much point in having this separate object. The methods are fine to have in the pipeline-builder.

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaUtils.scala
```diff
@@ -0,0 +1,119 @@
...
+  implicit val msgCoder: Coder[Msg] = Coder.beam(new UpackMsgCoder)
+  implicit val readableFileCoder: Coder[ReadableFile] = Coder.beam(new ReadableFileCoder)
+
+  // format is: metadata/{entity_type}/{entity_id}_{version}.json,
+  // but filename returns just {entity_id}_{version}.json, so that is what we deal with.
+  val metadataPattern: Regex = "([\\S]+)_([\\S]+).json".r
```
danxmoran, line 23: Instead of `[\\S]+`, could the 1st one be `[^_]+` and the 2nd just be `.+`?

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaUtils.scala
```diff
@@ -0,0 +1,119 @@
...
+    *
+    * extract the necessary info from a metadata file and put it into a form that makes it easy to
+    * pass in to the table format
+    *
+    * @param entityType the metadata entity type to prepend for the id field
+    * @param fileName the raw filename of the metadata file
+    * @param metadata the content of the metadata file in Msg format
+    * @return a Msg object in the desired output format
+    */
+  def transformMetadata(entityType: String, fileName: String, metadata: Msg): Msg = {
```
danxmoran, line 66: Would it be worth defining a `case class` to return here instead of leaving things as `Msg`?

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaUtils.scala
```diff
@@ -0,0 +1,119 @@
...
+        )
+      )
+    val entityId = matches.group(1)
+    val entityVersion = matches.group(2)
+    // and put in form we want
+    Obj(
+      mutable.LinkedHashMap[Msg, Msg](
+        Str(s"${entityType}_id") -> Str(entityId),
+        Str("version") -> Str(entityVersion),
+        Str("content") -> metadata
```
danxmoran, line 82: `metadata` needs to be stringified. See how ClinVar [does it](https://github.com/DataBiosphere/clinvar-ingest/blob/master/transformation/src/main/scala/org/broadinstitute/monster/clinvar/Content.scala#L17).

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaUtils.scala
```diff
@@ -0,0 +1,119 @@
...
+    entityType: String
+  ): ClosedTap[String] = {
+    // get the readable files for the given input path
+    val readableFiles = getReadableFiles(
+      s"$inputPrefix/metadata/${entityType}/**.json",
+      context
+    )
+    // then convert json to msg and get the filename
+    val processedData = jsonToFilenameAndMsg(entityType)(readableFiles)
+      .withName(s"Convert ${entityType} from JSON to Msg in output form")
```
danxmoran, line 108: Name here is pretty low-level, suggested alternative is:
```suggestion
      .withName(s"Pre-process ${entityType} metadata")
```

build.sbt
```diff
@@ -0,0 +1,47 @@
+//import _root_.io.circe.Json
+import org.broadinstitute.monster.sbt.model.JadeIdentifier
+
+val betterFilesVersion = "3.8.0"
```
danxmoran, line 4: Are we using this anywhere?

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaUtils.scala
raaidbroad: Would that work with every table having a differently named `id` column? If so, sure thing

danxmoran: Oh right, ugh. This is ok then 👍

PR #4, opened 2020-05-22
Sw dspdc 1082 unit test for metadata transformation

snwessel:
Writing a unit test for the metadataTransformation method ([Jira Ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1082)).

I had been debating writing a test for the processMetadata method too, but wasn't sure if it was worth the effort of setting up the file test input/output files and temporary directory. If anyone feels like we should have it, I'd be happy to add it in.

### DIFF
--- a/transformation/src/test/scala/org/broadinstitute/monster/hca/HcaPipelineBuilderSpec.scala
+++ b/transformation/src/test/scala/org/broadinstitute/monster/hca/HcaPipelineBuilderSpec.scala
@@ -7,17 +7,17 @@ import org.scalatest.matchers.should.Matchers
 class HcaPipelineBuilderSpec extends AnyFlatSpec with Matchers {
   behavior of "HcaPipelineBuilder"
 
-  it should "transform metadata" in {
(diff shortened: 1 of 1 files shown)


transformation/src/test/scala/org/broadinstitute/monster/hca/HcaPipelineBuilderSpec.scala
```diff
@@ -0,0 +1,29 @@
...
+    val exampleFileContent = JsonParser.parseEncodedJson("""{"beep": "boop"}""")
+    val actualOutput = HcaPipelineBuilder.transformMetadata(
+      entityType = "entity_type",
+      fileName = "entityId_version.json",
+      metadata = exampleFileContent
+    )
+    val expectedOutput = JsonParser.parseEncodedJson(
+      json = """
+               | {
+               |   "entity_type_id": "entityId",
```
quazi-broad, line 20: This looks consistent with the logic in HcaPipelineBuilder.scala.
LGTM!

PR #6, opened 2020-05-28
Sw dspdc 1077 process file metadata

snwessel:
Transforming the file metadata: 
- file_id and file_version are parsed out of the name of the data file
- the entity id and version are parsed out of the name of the metadata file (the same way they are for the rest of the metadata)
- the checksum value is read from the json

### DIFF
--- a/transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
+++ b/transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
@@ -34,7 +34,7 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
   // format is: metadata/{entity_type}/{entity_id}_{version}.json,
   // but filename returns just {entity_id}_{version}.json, so that is what we deal with.
   val metadataPattern: Regex = "([^_]+)_(.+).json".r
-  // format is: {dir_path}{file_id}_{file_version}_{file_name},
+  // format is: {dir_path}/{file_id}_{file_version}_{file_name},
   // but the dir_path seems to be optional
   val fileDataPattern: Regex = "(.*\\/)?([^_^\\/]+)_([^_]+)_(.+)".r
 
@@ -145,7 +145,7 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
         Str("content") -> Str(encode(metadata).getOrElse("")),
         Str("file_id") -> Str(fileId),
         Str("file_version") -> Str(fileVersion),
-        Str("content_hash") -> Str(checksum.getOrElse(""))
+        Str("checksum") -> Str(checksum.getOrElse(""))
       )
     )
   }
--- a/transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
+++ b/transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
@@ -18,12 +18,11 @@ import scala.util.matching.Regex
 object HcaPipelineBuilder extends PipelineBuilder[Args] {
 
   override def buildPipeline(ctx: ScioContext, args: Args): Unit = {
-    // process metadata for each non-file metadata entity
-    metadataEntities.foreach { entityType =>
-      processMetadata(ctx, args.inputPrefix, args.outputPrefix, entityType)
-    }
-    fileMetadataEntities.foreach { entityType =>
-      processMetadata(ctx, args.inputPrefix, args.outputPrefix, entityType, isFileMetadata = true)
+    val allMetadataEntities =
+      metadataEntities.map(_ -> false) ++ fileMetadataEntities.map(_ -> true)
+    allMetadataEntities.foreach {
+      case (entityType, isFileMetadata) =>
+        processMetadata(ctx, args.inputPrefix, args.outputPrefix, entityType, isFileMetadata)
     }
     ()
   }
@@ -113,7 +112,7 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
     * @return a Msg object in the desired output format
     */
   def transformMetadata(entityType: String, fileName: String, metadata: Msg): Msg = {
-    val (entityId, entityVersion) = getEntityIdAndVersion(fileName: String)
+    val (entityId, entityVersion) = getEntityIdAndVersion(fileName)
     Obj(
       mutable.LinkedHashMap[Msg, Msg](
         Str(s"${entityType}_id") -> Str(entityId),
@@ -134,8 +133,9 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
     * @return a Msg object in the desired output format
     */
   def transformFileMetadata(entityType: String, fileName: String, metadata: Msg): Msg = {
-    val (entityId, entityVersion) = getEntityIdAndVersion(fileName: String)
-    val dataFileName = metadata.read[Msg]("file_core").read[String]("file_name")
+    val (entityId, entityVersion) = getEntityIdAndVersion(fileName)
+    val contentHash = metadata.read[String]("file_core", "file_crc32c")
+    val dataFileName = metadata.read[String]("file_core", "file_name")
     val (fileId, fileVersion) = getFileIdAndVersion(dataFileName)
     // put values in the form we want
     Obj(
@@ -143,6 +143,7 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
         Str(s"${entityType}_id") -> Str(entityId),
         Str("version") -> Str(entityVersion),
         Str("content") -> Str(encode(metadata).getOrElse("")),
+        Str("content_hash") -> Str(contentHash),
         Str("source_file_id") -> Str(fileId),
         Str("source_file_version") -> Str(fileVersion)
       )
--- a/transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
+++ b/transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
@@ -134,10 +134,8 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
     * @return a Msg object in the desired output format
     */
   def transformFileMetadata(entityType: String, fileName: String, metadata: Msg): Msg = {
-    val coreFileMetadata = metadata.read[Msg]("file_core")
-    val checksum = coreFileMetadata.tryRead[String]("checksum")
-    val dataFileName = coreFileMetadata.read[String]("file_name")
     val (entityId, entityVersion) = getEntityIdAndVersion(fileName: String)
+    val dataFileName = metadata.read[Msg]("file_core").read[String]("file_name")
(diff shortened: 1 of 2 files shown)


snwessel:
Just looked back through the design doc and realized I need to update a few more things... so hold off on reviewing this for the moment

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
```diff
@@ -99,24 +113,78 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
...
+    val dataFileName = metadata.read[Msg]("file_core").read[String]("file_name")
+    val (fileId, fileVersion) = getFileIdAndVersion(dataFileName)
+    // put values in the form we want
+    Obj(
+      mutable.LinkedHashMap[Msg, Msg](
+        Str(s"${entityType}_id") -> Str(entityId),
+        Str("version") -> Str(entityVersion),
+        Str("content") -> Str(encode(metadata).getOrElse("")),
+        Str("source_file_id") -> Str(fileId),
+        Str("source_file_version") -> Str(fileVersion)
```
snwessel, line 147: We aren't extracting the checksum field for now, right? The [importer design doc](https://docs.google.com/document/d/1AJfpaGMauFlusgAOKQHNeCta-PrHz5OObO-FcHyn_zw/edit?usp=sharing) seemed to suggest that we would hold off on using it for now, and the file metadata I looked through just had the checksum field set to null.

I wanted to confirm because the description on the [jira ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1077) suggested that we _would_ want to use some kind of content hash.

snwessel:
Okay, ready for review now 👍

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
```diff
@@ -31,6 +34,9 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
   // format is: metadata/{entity_type}/{entity_id}_{version}.json,
   // but filename returns just {entity_id}_{version}.json, so that is what we deal with.
   val metadataPattern: Regex = "([^_]+)_(.+).json".r
+  // format is: {dir_path}/{file_id}_{file_version}_{file_name},
+  // but the dir_path seems to be optional
+  val fileDataPattern: Regex = "(.*\\/)?([^_^\\/]+)_([^_]+)_(.+)".r
```
danxmoran, line 39: https://xkcd.com/208/

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
danxmoran: I think we will need a checksum in both these outputs and the bulk-file request outputs. They won't have a column in BQ, but they're needed for a BQ-based diffing pre-step.

Are the checksums still null after UCSC refreshed their test bucket yesterday? If so, we should ask them about it.

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
```diff
@@ -22,6 +22,9 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
     metadataEntities.foreach { entityType =>
       processMetadata(ctx, args.inputPrefix, args.outputPrefix, entityType)
     }
+    fileMetadataEntities.foreach { entityType =>
+      processMetadata(ctx, args.inputPrefix, args.outputPrefix, entityType, isFileMetadata = true)
+    }
```
danxmoran, line 27: The way this is set up, we could have a single list of entities of type `List[(String, Boolean}]` and then only run a single `foreach` over it.

danxmoran:
```scala
metadataEntities.map(_ -> false) ++ fileMetadataEntities.map(_ -> true)
```
Would save us from typing out a bunch of hard-coded bools

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
```diff
@@ -99,24 +113,78 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
     * @return a Msg object in the desired output format
     */
   def transformMetadata(entityType: String, fileName: String, metadata: Msg): Msg = {
-    // extract info from file name
+    val (entityId, entityVersion) = getEntityIdAndVersion(fileName: String)
```
danxmoran, line 116: I don't think the type assertion is needed in the argument list
```suggestion
    val (entityId, entityVersion) = getEntityIdAndVersion(fileName)
```

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
```diff
@@ -99,24 +113,78 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
...
+    * Extract the necessary info from a file of file-related metadata and put it into a form that
+    * makes it easy to pass in to the table format
+    *
+    * @param entityType the metadata entity type to prepend for the id field
+    * @param fileName the raw filename of the metadata file
+    * @param metadata the content of the metadata file in Msg format
+    * @return a Msg object in the desired output format
+    */
+  def transformFileMetadata(entityType: String, fileName: String, metadata: Msg): Msg = {
+    val (entityId, entityVersion) = getEntityIdAndVersion(fileName: String)
```
danxmoran, line 137:
```suggestion
    val (entityId, entityVersion) = getEntityIdAndVersion(fileName)
```

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
```diff
@@ -99,24 +113,78 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
...
+    val dataFileName = metadata.read[Msg]("file_core").read[String]("file_name")
```
danxmoran, line 138: I think this should work:
```suggestion
    val dataFileName = metadata.read[Msg]("file_core", "file_name")
```

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
```diff
@@ -99,24 +113,78 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
     * @return a Msg object in the desired output format
     */
   def transformMetadata(entityType: String, fileName: String, metadata: Msg): Msg = {
-    // extract info from file name
+    val (entityId, entityVersion) = getEntityIdAndVersion(fileName: String)
+    Obj(
+      mutable.LinkedHashMap[Msg, Msg](
+        Str(s"${entityType}_id") -> Str(entityId),
+        Str("version") -> Str(entityVersion),
+        Str("content") -> Str(encode(metadata).getOrElse(""))
```
danxmoran, line 121: Oof, my mistake for missing this in the original code. `getOrElse("")` is bad (it hides a lack of data). We should either:
1. Throw an exception here if `content` is truly required, or
2. Have `content` be nullable, and leave it out when `encode(metadata)` returns `None`

Under what circumstances would we expect `encode` to return `None`?

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
```diff
@@ -99,24 +113,78 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
...
+    (entityId, entityVersion)
+  }
+
+  /**
+    * Extract the file id and file version from the name of a data file.
+    *
+    * @param fileName the raw filename of the data file
+    * @return a tuple of the file id and file version
+    */
+  def getFileIdAndVersion(fileName: String): (String, String) = {
```
danxmoran, line 177: This is good for now, but I expect we'll want to also return the directory path as part of the follow-on ticket

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
snwessel: Yeah, the checksum field is still null. Before the refresh, it wasn't a part of the json file at all, now the field is listed but set to: `"checksum": null`.

I'll update my changes to use it though 👍

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
snwessel: lol that's great 😆 we're all just waiting our moments to shine...

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
snwessel: Wow 🤔 I really don't know why I did that! Thanks for catching it though!

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
snwessel: Then `encode` method only returns none when the metadata Msg is empty. Would it make sense to update the `encode` method itself to throw an exception when its input is empty, and then change its output type to be a string (instead of an `Option[String]`)?

danxmoran: In this case, I don't think we need to check emptiness of the input object at all. If it's an empty object then the desired output is `"{}"`

snwessel: Sounds good. I just updated that method and then wrote another quick test to make sure it was actually outputing the `{}`.

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
```diff
@@ -99,32 +112,85 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
...
+    val contentHash = metadata.read[String]("file_core", "file_crc32c")
+    val dataFileName = metadata.read[String]("file_core", "file_name")
+    val (fileId, fileVersion) = getFileIdAndVersion(dataFileName)
+    // put values in the form we want
+    Obj(
+      mutable.LinkedHashMap[Msg, Msg](
+        Str(s"${entityType}_id") -> Str(entityId),
+        Str("version") -> Str(entityVersion),
+        Str("content") -> Str(encode(metadata)),
+        Str("content_hash") -> Str(contentHash),
```
danxmoran, line 146: Let's name this more explicitly so it's clear which hash it contains:
```suggestion
        Str("crc32c") -> Str(contentHash),
```

PR #7, opened 2020-05-28
Ra dspdc 1053 init argo

raaidbroad:
Initialize argo workflow for the hca ingest.


orchestration/Chart.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v2
+name: hca-ingest
+description: Top-level orchestration workflow for HCA
+type: application
+# TODO: This needs to auto-update on every merge
```
danxmoran, line 5: This TODO is now implemented in the sbt plugin, and could be removed.

orchestration/templates/ingest-hca.yaml
```diff
@@ -0,0 +1,61 @@
...
+          - name: source-bucket-name
+          - name: staging-bucket-name
+
+      dag:
+        tasks:
+          # Pre-process the data found in a source bucket
+          - name: pre-process-source
+            templateRef:
+              name: run-dataflow
+              template: main
```
danxmoran, line 21: Since it's in the same file, it should just be:
```
template: run-dataflow
```
instead of the `templateRef` block

orchestration/templates/ingest-hca.yaml
```diff
@@ -0,0 +1,61 @@
...
+    - name: main
+      inputs:
+        parameters:
+          - name: source-bucket-name
+          - name: staging-bucket-name
+
+      dag:
+        tasks:
+          # Pre-process the data found in a source bucket
+          - name: pre-process-source
```
danxmoran, line 18:
```suggestion
          - name: pre-process-metadata
```
For consistency with their terminology

orchestration/templates/ingest-hca.yaml
```diff
@@ -0,0 +1,61 @@
...
+  name: ingest-hca
+spec:
+  entrypoint: main
+  serviceAccountName: {{ .Values.serviceAccount.k8sName }}
+  templates:
+    - name: main
+      inputs:
+        parameters:
+          - name: source-bucket-name
+          - name: staging-bucket-name
```
danxmoran, line 13: Could we make this a Helm variable?

orchestration/templates/ingest-hca.yaml
```diff
@@ -0,0 +1,61 @@
...
+metadata:
+  name: ingest-hca
+spec:
+  entrypoint: main
+  serviceAccountName: {{ .Values.serviceAccount.k8sName }}
+  templates:
+    - name: main
+      inputs:
+        parameters:
+          - name: source-bucket-name
```
danxmoran, line 12: I think we'll need a `prefix` along with the bucket name

orchestration/templates/ingest-hca.yaml
```diff
@@ -0,0 +1,61 @@
...
+                - name: staging-bucket-name
+                  value: '{{ "{{inputs.parameters.staging-bucket-name}}" }}'
+
+    # Template used to launch a Dataflow processing job on raw ClinVar data,
+    # transforming it to our target schema.
+    - name: run-dataflow
+      inputs:
+        parameters:
+          - name: source-bucket-name
+          - name: staging-bucket-name
```
danxmoran, line 35: If we decide to make this a Helm var, would remove this parameter and inline the Helm lookup where it's used below.

orchestration/templates/ingest-hca.yaml
```diff
@@ -0,0 +1,61 @@
...
+          - --autoscalingAlgorithm=THROUGHPUT_BASED
+          - --numWorkers={{ .minWorkers }}
+          - --maxNumWorkers={{ .maxWorkers }}
+          {{- end }}
+          {{- if .useFlexRS }}
+          - --flexRSGoal=COST_OPTIMIZED
+          {{- else }}
+          - --experiments=shuffle_mode=service
+          {{- end }}
+          {{- end }}
```
danxmoran, line 61: Can you add the newline to EOF?

orchestration/values.schema.json
```diff
@@ -0,0 +1,60 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "properties": {
+    "gcs": {
+      "type": "object",
+      "properties": {
+        "sourceBucketName": { "type": "string" },
```
danxmoran, line 8: This doesn't need to be Helm, I expect it'll vary from run to run.

orchestration/templates/ingest-hca.yaml
danxmoran: Ah, it's already in the Helm schema, we just need to use it.

orchestration/values.schema.json
```diff
@@ -0,0 +1,60 @@
...
+    },
+    "serviceAccount": {
+      "type": "object",
+      "properties": {
+        "k8sName": { "type": "string" },
+        "googleName": { "type": "string" }
+      },
+      "required": ["k8sName", "googleName"]
+    },
+    "cron": {
```
danxmoran, line 21: Don't need `cron` for this one, since we won't be running it on a schedule

orchestration/values.yaml
```diff
@@ -0,0 +1,12 @@
+cron:
+  enable: true
+  # Every Sunday at 3:30am
+  schedule: '30 3 * * Sun'
+  timezone: 'America/New_York'
+dataflow:
+  region: us-central1
+  workerMachineType: n1-standard-4
```
danxmoran, line 8: We can probably go wimpy here, would default to `n1-standard-1`

orchestration/values.yaml
```diff
@@ -0,0 +1,12 @@
...
+  autoscaling:
+    minWorkers: 4
+    maxWorkers: 8
+  useFlexRS: false
```
danxmoran, line 12: Can you add a newline to EOF?

orchestration/templates/ingest-hca.yaml
raaidbroad: Given that each bucket is supposed to fit a particular layout (where it has the metadata, data, and links directories), what would the prefix be used for?

danxmoran: The prefix gives more flexibility in storing multiple "areas" within a single bucket (i.e. EBI wants to use one bucket for both dev and prod). metadata/data/links are all relative to bucket + prefix.

orchestration/example-values/no-version.yaml
```diff
@@ -0,0 +1,16 @@
+gcs:
+  stagingBucketName: buc
```
danxmoran, line 2: Excellent

orchestration/templates/ingest-hca.yaml
```diff
@@ -0,0 +1,61 @@
...
+  name: ingest-hca
+spec:
+  entrypoint: main
+  serviceAccountName: {{ .Values.serviceAccount.k8sName }}
+  templates:
+    - name: main
+      inputs:
+        parameters:
+          - name: source-bucket-name
+          - name: source-bucket-prefix
```
danxmoran, line 13: Sorry, one last thing: I think it'd be good to include an `output-prefix` as an arg here, and pass it through to the dataflow job. That way we can reprocess the same inputs multiple times without the outputs colliding.

orchestration/templates/ingest-hca.yaml
```diff
@@ -0,0 +1,61 @@
...
+          {{- $sourcePrefix := "{{inputs.parameters.source-bucket-prefix}}" }}
+      container:
+        {{- $version := default "latest" .Chart.AppVersion }}
+        image: us.gcr.io/broad-dsp-gcr-public/hca-transformation-pipeline:{{ $version }}
+        command: []
+        args:
+          - --runner=dataflow
+          - --inputPrefix=gs://{{ $sourceBucket }}/{{ $sourcePrefix }}
+          - --outputPrefix=gs://{{ .Values.gcs.stagingBucketName }}
```
danxmoran, line 43: This would then be `gs://staging-bucket/output-prefix`

PR #8, opened 2020-05-29
Sw dspdc 1078 generate file ingest requests

snwessel:
Generating a file ingest request for each file with the following values:
- the file's source path
- the file's target path
- a crc32c hash of the file's contents

@danxmoran I wasn't sure what you had in mind as the input for the method to create file ingest requests. I am using the pre-processed metadata as the input here, and added the "data file name" as a field on the pre-processed data so it could be passed in to this new method. If you prefer that I use the raw metadata as an input instead (to avoid messing with the `transformFileMetadata ` method), I would be happy to do that. Let me know what you think.

### DIFF
--- a/transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
+++ b/transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
@@ -18,12 +18,14 @@ import scala.util.matching.Regex
 object HcaPipelineBuilder extends PipelineBuilder[Args] {
 
   override def buildPipeline(ctx: ScioContext, args: Args): Unit = {
+    // transform metadata
     val allMetadataEntities =
       metadataEntities.map(_ -> false) ++ fileMetadataEntities.map(_ -> true)
     allMetadataEntities.foreach {
       case (entityType, isFileMetadata) =>
         processMetadata(ctx, args.inputPrefix, args.outputPrefix, entityType, isFileMetadata)
     }
+    // generateFileIngestRequests
     ()
   }
 
@@ -145,7 +147,8 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
         Str("content") -> Str(encode(metadata)),
         Str("crc32c") -> Str(contentHash),
         Str("source_file_id") -> Str(fileId),
-        Str("source_file_version") -> Str(fileVersion)
+        Str("source_file_version") -> Str(fileVersion),
+        Str("data_file_name") -> Str(dataFileName)
       )
     )
   }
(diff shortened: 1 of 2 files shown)


transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
```diff
@@ -220,6 +243,15 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
           if (isFileMetadata) transformFileMetadata(entityType, filename, metadata)
           else transformMetadata(entityType, filename, metadata)
       }
+    // generate a file ingest request (if applicable)
+    if (isFileMetadata) {
+      val fileIngestRequests = processedData.map(generateFileIngestRequest(_, inputPrefix))
+      StorageIO.writeJsonLists(
+        fileIngestRequests,
+        entityType,
+        s"${outputPrefix}/data/${entityType}"
```
danxmoran, line 252: I'd use something other than `data`, since that already has a meaning in the common layout. Maybe `data-transfer-requests`?

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
```diff
@@ -145,7 +145,30 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
...
+    * can be used to generate bulk file ingest requests.
+    * @param metadata the content of the metadata file in Msg format
+    * @param inputPrefix the root directory containing input files
+    * @return a Msg object in the desired output format
+    */
+  def generateFileIngestRequest(metadata: Msg, inputPrefix: String): Msg = {
+    val contentHash = metadata.read[String]("crc32c")
+    val dataFileName = metadata.read[String]("data_file_name")
+    val sourcePath = s"$inputPrefix/data/$dataFileName"
+    val targetPath = s"/$dataFileName"
```
danxmoran, line 165: Sanity-checking: Are we sure the data file names don't start with `/`?

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
snwessel: Sounds good! I just updated it.

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
snwessel: I just checked, and no, they don't start with `/`.

PR #13, opened 2020-06-08
DSPDC-1069 Run bulk-file ingest after diffing requests

danxmoran:
Apart from the bug in the repo, this is working perfectly.


danxmoran:
Here's an example where the steps in this PR succeeded (even though the downstream steps failed... still working on those): https://hca-argo.monster-dev.broadinstitute.org/archived-workflows/hca-mvp/422199a3-4028-4a50-b6fc-ed2ca0fade39

PR #18, opened 2020-06-26
Ra dspdc 1136 new checksums

raaidbroad:
Update the file metadata transformation pieces so that they use the new descriptor files. Also updates the schema for tables that end in "*_file" and updates relevant tests. Deleted one test case that doesn't apply anymore.

### DIFF
--- a/schema/src/main/jade-tables/analysis_file.table.json
+++ b/schema/src/main/jade-tables/analysis_file.table.json
@@ -5,8 +5,6 @@
     { "name": "version", "datatype": "timestamp", "type": "primary_key" },
     { "name": "content", "datatype": "string", "type": "required" },
     { "name": "file_id", "datatype": "fileref", "type": "required" },
-    { "name": "source_file_id", "datatype": "string", "type": "required" },
-    { "name": "source_file_version", "datatype": "timestamp", "type": "required" },
     { "name": "descriptor", "datatype": "string", "type": "required" }
   ],
   "partitionMode": "date",
--- a/schema/src/main/jade-tables/analysis_file.table.json
+++ b/schema/src/main/jade-tables/analysis_file.table.json
@@ -6,7 +6,8 @@
     { "name": "content", "datatype": "string", "type": "required" },
     { "name": "file_id", "datatype": "fileref", "type": "required" },
     { "name": "source_file_id", "datatype": "string", "type": "required" },
-    { "name": "source_file_version", "datatype": "timestamp", "type": "required" }
+    { "name": "source_file_version", "datatype": "timestamp", "type": "required" },
+    { "name": "descriptor", "datatype": "string", "type": "required" }
   ],
   "partitionMode": "date",
   "datePartitionOptions": {
--- a/transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
+++ b/transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
@@ -154,7 +154,7 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
         Str("version") -> Str(entityVersion),
         Str("content") -> Str(encode(metadata)),
         Str("virtual_path") -> Str(s"/$dataFileName"),
-        Str("crc32c") -> Str(descriptor.read[String]("crc32c")),
+        Str("crc32c") -> descriptor.read[Msg]("crc32c"),
         Str("descriptor") -> Str(encode(descriptor))
       )
     )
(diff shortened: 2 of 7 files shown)


schema/src/main/jade-tables/analysis_file.table.json
```diff
@@ -6,7 +6,8 @@
     { "name": "content", "datatype": "string", "type": "required" },
     { "name": "file_id", "datatype": "fileref", "type": "required" },
     { "name": "source_file_id", "datatype": "string", "type": "required" },
-    { "name": "source_file_version", "datatype": "timestamp", "type": "required" }
+    { "name": "source_file_version", "datatype": "timestamp", "type": "required" },
```
danxmoran, line 9: Related change that landed yesterday: We don't need to track `source_file_id` or `source_file_version` now that the info is included in `descriptor`, so we can delete those columns across the board.

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
```diff
@@ -281,21 +287,31 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
...
+      metadataFilenameAndMsg.join(descriptorFilenameAndMsg)
+    } else {
+      // for non-file metadata
+      jsonToFilenameAndMsg(entityType)(metadataFiles)
+    }
+
+    val processedMetadata = readyMetadata.withName(s"Pre-process $entityType metadata").map {
+      case (filename: String, metadata: Msg) => transformMetadata(entityType, filename, metadata)
+      case (filename: String, (metadata: Msg, descriptor: Msg)) =>
+        transformFileMetadata(entityType, filename, metadata, descriptor)
```
danxmoran, line 309: Since these are totally-divergent paths, would prefer to push these transformation calls into the `if` / `else` above.

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
```diff
@@ -281,21 +287,31 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
...
+    val readyMetadata = if (isFileMetadata) {
+      val descriptorFiles = getReadableFiles(
+        s"$inputPrefix/descriptors/$entityType/**.json",
+        context
+      )
+      val descriptorFilenameAndMsg = jsonToFilenameAndMsg(entityType)(descriptorFiles)
+      metadataFilenameAndMsg.join(descriptorFilenameAndMsg)
+    } else {
+      // for non-file metadata
+      jsonToFilenameAndMsg(entityType)(metadataFiles)
```
danxmoran, line 303: This could be `metadataFilenameAndMsg`, yes?

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
```diff
@@ -281,21 +287,31 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
...
-      .map {
-        case (filename, metadata) =>
-          if (isFileMetadata) transformFileMetadata(entityType, filename, metadata)
-          else transformMetadata(entityType, filename, metadata)
-      }
+    val metadataFiles = getReadableFiles(s"$inputPrefix/metadata/${entityType}/**.json", context)
+    val metadataFilenameAndMsg = jsonToFilenameAndMsg(entityType)(metadataFiles)
+
+    // for file metadata
+    val readyMetadata = if (isFileMetadata) {
```
danxmoran, line 294: Would be nice if we could unit-test from here down (`ReadableFile` is declared `final` so I don't think we can mock it to include `jsonToFilenameAndMsg` in a unit-test).

Could you think through what it might look like to shift this business logic into a separate class / method? Ideally it'd be unit-testable independently from the implementations of `transformMetadata` and `transformFileMetadata` (we'd need some interfaces in front of those which we could mock).

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
```diff
@@ -136,25 +136,26 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
...
       mutable.LinkedHashMap[Msg, Msg](
         Str(s"${entityType}_id") -> Str(entityId),
         Str("version") -> Str(entityVersion),
         Str("content") -> Str(encode(metadata)),
-        Str("crc32c") -> Str(contentHash),
-        Str("source_file_id") -> Str(fileId),
-        Str("source_file_version") -> Str(fileVersion),
-        Str("virtual_path") -> Str(s"/$dataFileName")
+        Str("virtual_path") -> Str(s"/$dataFileName"),
+        Str("crc32c") -> Str(descriptor.read[String]("crc32c")),
```
danxmoran, line 157:
```suggestion
        Str("crc32c") -> descriptor.read[Msg]("crc32c"),
```
Would pass through whatever value is stored in the field, with less object allocation / parsing. It wouldn't assert that the value is a string, though.

PR #19, opened 2020-06-30
DSPDC-1157 Fix partitioning spec.

danxmoran:
Schema JSONs were using Jade-style syntax, not our own model. Example successful run is here: https://hca-argo.monster-dev.broadinstitute.org/workflows/hca-mvp/ingest-hca-kjgv7-cbrzh-s6rpj-42p9l-7stsw-rgf55

### DIFF
--- a/transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
+++ b/transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
@@ -12,7 +12,6 @@ import org.broadinstitute.monster.common.msg._
 import ujson.StringRenderer
 import upack.{Msg, Obj, Str}
 
-import scala.collection.mutable
 import scala.util.matching.Regex
 
 object HcaPipelineBuilder extends PipelineBuilder[Args] {
@@ -120,11 +119,9 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
   def transformMetadata(entityType: String, fileName: String, metadata: Msg): Msg = {
     val (entityId, entityVersion) = getEntityIdAndVersion(fileName)
     Obj(
-      mutable.LinkedHashMap[Msg, Msg](
-        Str(s"${entityType}_id") -> Str(entityId),
-        Str("version") -> Str(entityVersion),
-        Str("content") -> Str(encode(metadata))
-      )
+      Str(s"${entityType}_id") -> Str(entityId),
+      Str("version") -> Str(entityVersion),
+      Str("content") -> Str(encode(metadata))
     )
   }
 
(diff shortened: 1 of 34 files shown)


transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
```diff
@@ -146,17 +143,13 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
...
   ): Msg = {
     val (entityId, entityVersion) = getEntityIdAndVersion(fileName)
-    val dataFileName = descriptor.read[String]("file_name")
     // put values in the form we want
     Obj(
-      mutable.LinkedHashMap[Msg, Msg](
-        Str(s"${entityType}_id") -> Str(entityId),
-        Str("version") -> Str(entityVersion),
-        Str("content") -> Str(encode(metadata)),
-        Str("virtual_path") -> Str(s"/$dataFileName"),
```
raaidbroad, line 156: did you intend to drop the virtual_path?

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
```diff
@@ -173,33 +166,31 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
...
+  ): (String, Msg) = {
+    val contentHash = descriptor.read[String]("crc32c")
+    val targetPath = descriptor.read[String]("file_name")
+    val sourcePath = s"$inputPrefix/data/$targetPath"
 
-    Obj(
-      mutable.LinkedHashMap[Msg, Msg](
-        Str("source_path") -> Str(sourcePath),
-        Str("target_path") -> Str(targetPath),
-        Str("crc32c") -> Str(contentHash)
```
raaidbroad, line 201: did you intend to drop the crc32c?

transformation/src/test/scala/org/broadinstitute/monster/hca/HcaPipelineBuilderSpec.scala
```diff
@@ -86,8 +86,7 @@ class HcaPipelineBuilderSpec extends AnyFlatSpec with Matchers {
           |   "version": "entity-version",
           |   "content": "{\"file_core\":{\"file_name\":\"some-id_some-version.numbers123_12-34_metrics_are_fun.csv\",\"format\":\"csv\",\"file_provenance\":{\"crc32c\":\"54321zyx\"}},\"schema_type\":\"file\"}",
           |   "crc32c": "54321zyx",
-          |   "descriptor": "{\"file_name\":\"some-id_some-version.numbers123_12-34_metrics_are_fun.csv\",\"file_id\":\"my-file-id\",\"file_version\":\"my-file-version\",\"crc32c\":\"54321zyx\",\"schema_type\":\"file_descriptor\"}",
-          |   "virtual_path": "/some-id_some-version.numbers123_12-34_metrics_are_fun.csv"
```
raaidbroad, line 90: guessing we are dropping that virtual path then haha

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
quazi-broad: oh, actually, I think that was moved to the line above (187)
`val contentHash = descriptor.read[String]("crc32c")`

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
danxmoran: Yes. Instead of using the file metadata to generate the bulk request, we use the raw descriptor. So there's no need for these extra top-level fields.

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
danxmoran: Yes. It's still in the descriptor JSON, and we can extract it using BigQuery's JSON functions.

PR #22, opened 2020-07-16
Ra dspdc 1128 schema validation

raaidbroad:
Add json schema validation to our piece of the HCA pipeline. Between the additional Circe JSON encoding and the requests-scala library, I don't think this is very high performance, but it sure seems to work.

### DIFF
--- a/transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
+++ b/transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
@@ -1,5 +1,6 @@
 package org.broadinstitute.monster.hca
 
+import cats.data.Validated
 import com.spotify.scio.ScioContext
 import com.spotify.scio.coders.Coder
 import com.spotify.scio.io.ClosedTap
@@ -261,9 +262,9 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
   def encode(msg: Msg): String =
     upack.transform(msg, StringRenderer()).toString
 
-  def validateJson(value: SCollection[(String, Msg)]): Unit = {
+  def validateJson(filenamesAndMsg: SCollection[(String, Msg)]): Unit = {
     // pull out the url for where the schema definition is for each file
-    val content = value.map {
+    val content = filenamesAndMsg.map {
       case (filename, msg) => (msg.read[String]("describedBy"), (filename, msg))
     }
     // get the distinct urls (so as to minimize the number of get requests) and then get the schemas as strings
@@ -284,12 +285,20 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
                 )
               case Success(value) =>
                 // try to parse the actual data into a json format for validation
-                parse(data.str) match {
+                parse(encode(data)) match {
                   case Left(_) =>
                     throw new Exception(s"Unable to parse data into json for file $filename")
                   // if everything is parsed/encoded/etc correctly, actually try to validate against schema here
                   // if not valid, will return list of issues
-                  case Right(success) => value.validate(success)
+                  case Right(success) => {
+                    value.validate(success) match {
+                      case Validated.Valid(_) => ()
+                      case Validated.Invalid(e) =>
+                        throw new Exception(
+                          s"Data does not conform to schema: ${e.map(_.getMessage)}"
+                        )
+                    }
+                  }
                 }
             }
           }
--- a/transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
+++ b/transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
@@ -6,7 +6,6 @@ import com.spotify.scio.coders.Coder
 import com.spotify.scio.io.ClosedTap
 import com.spotify.scio.values.SCollection
 
-import scala.util.{Failure, Success}
 import io.circe.parser._
 import io.circe.schema.Schema
 import org.apache.beam.sdk.io.{FileIO, ReadableFileCoder}
@@ -17,6 +16,7 @@ import org.broadinstitute.monster.common.msg._
 import ujson.StringRenderer
 import upack.{Msg, Obj, Str}
 
+import scala.util.{Failure, Success}
 import scala.util.matching.Regex
 
 object HcaPipelineBuilder extends PipelineBuilder[Args] {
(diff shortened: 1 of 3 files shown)


transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
```diff
@@ -1,9 +1,14 @@
 package org.broadinstitute.monster.hca
 
+import cats.data.Validated
 import com.spotify.scio.ScioContext
 import com.spotify.scio.coders.Coder
 import com.spotify.scio.io.ClosedTap
 import com.spotify.scio.values.SCollection
+
+import scala.util.{Failure, Success}
```
danxmoran, line 9: Nitpick: `scala.*` imports are typically grouped together at the bottom of the list. I've tried to automated it but scalafmt isn't smort enough yet.

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
```diff
@@ -275,6 +328,8 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
     val metadataFiles = getReadableFiles(s"$inputPrefix/metadata/${entityType}/**.json", context)
     val metadataFilenameAndMsg = jsonToFilenameAndMsg(entityType)(metadataFiles)
 
+    validateJson(metadataFilenameAndMsg)
```
danxmoran, line 331: This will probably work in practice, but it isn't _technically_ the right control-flow. Calling this as a `Unit` method here registers a side-branch of processing in the graph. It's completely disconnected from the steps below, so when the pipeline is sent to Dataflow it'll be run in parallel with processing, instead of in sequence.

The fix is to have `validateJson` return an `SCollection`, and use that output as inputs downstream.

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
```diff
@@ -256,6 +262,53 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
   def encode(msg: Msg): String =
     upack.transform(msg, StringRenderer()).toString
 
+  def validateJson(filenamesAndMsg: SCollection[(String, Msg)]): Unit = {
```
danxmoran, line 265:
```suggestion
  def validateJson(filenamesAndMsg: SCollection[(String, Msg)]): SCollection[(String, Msg)] = {
```
See comment below in the main method for explanation

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
```diff
@@ -256,6 +262,53 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
...
+                    }
+                  }
+                }
+            }
+          }
+          case None => throw new Exception(s"No schema found at $url for file $filename")
+        }
+      }
+    }
+    ()
```
danxmoran, line 309:
```suggestion
```

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
```diff
@@ -256,6 +262,53 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
...
+              case Success(value) =>
+                // try to parse the actual data into a json format for validation
+                parse(encode(data)) match {
+                  case Left(_) =>
+                    throw new Exception(s"Unable to parse data into json for file $filename")
+                  // if everything is parsed/encoded/etc correctly, actually try to validate against schema here
+                  // if not valid, will return list of issues
+                  case Right(success) => {
+                    value.validate(success) match {
+                      case Validated.Valid(_) => ()
```
danxmoran, line 295:
```suggestion
                      case Validated.Valid(_) => (filename, data)
```

transformation/src/test/scala/org/broadinstitute/monster/hca/HcaPipelineBuilderSpec.scala
```diff
@@ -200,4 +211,83 @@ class HcaPipelineBuilderSpec extends AnyFlatSpec with Matchers {
...
+        |    "describedBy": "https://schema.humancellatlas.org/type/biomaterial/5.1.0/specimen_from_organism"
+        |}
+        |""".stripMargin
+    )
+
+    val exampleUrlAndFile = (exampleFileContent.read[String]("describedBy"), exampleFileContent)
+
+    the[Exception] thrownBy runWithContext { sc =>
+      HcaPipelineBuilder.validateJson(sc.parallelize(Seq(exampleUrlAndFile)))
+    } should have message "java.lang.Exception: Data does not conform to schema: NonEmptyList(#: required key [schema_type] not found)"
```
danxmoran, line 291: You could shorten this to be more flexible, in case there's any weird heuristics in the parser that might change the datatype used:
```suggestion
    } should include "Data does not conform to schema"
```

danxmoran: Oh, just noticed this is some magic helper for exceptions. Not sure if `include` works there.

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
```diff
@@ -256,6 +262,53 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
   def encode(msg: Msg): String =
     upack.transform(msg, StringRenderer()).toString
 
+  def validateJson(filenamesAndMsg: SCollection[(String, Msg)]): Unit = {
+    // pull out the url for where the schema definition is for each file
+    val content = filenamesAndMsg.map {
+      case (filename, msg) => (msg.read[String]("describedBy"), (filename, msg))
+    }
+    // get the distinct urls (so as to minimize the number of get requests) and then get the schemas as strings
+    val schemas = content.map(_._1).distinct.map(url => (url, requests.get(url).text))
```
danxmoran, line 271: OH, other thing to check: Try running this branch in Dataflow to make sure `requests` and the circe stuff are all serializable. If not, you'll get a runtime error.

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
raaidbroad: tfw the automation does it wrong

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
raaidbroad: ran it locally, it worked

transformation/src/test/scala/org/broadinstitute/monster/hca/HcaPipelineBuilderSpec.scala
raaidbroad: Yeah I was following an example from the scala testing docs

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
danxmoran: Can you link me to the dataflow job? So I can sanity-check the graph

PR #25, opened 2020-08-05
Sw dspdc 1199 schema validation logging

snwessel:
_Making this PR to get some general feedback on my approach before I finish implementing it. Any feedback is appreciated :) Let me know if there's a better way I could be doing this!_ 

Here are a couple of approaches I thought through:
- **My original (failed) approach:** Defining a new `SchemaValidationError` exception type and then trying to configure the logging in such a way that it writes a log whenever an exception of that type is thrown (haven't found a way to do this, so I gave up on it - let me know if you know of a way!)
- **My current (much simpler) approach**:  Updating the existing logger configurations to write any "ERROR" level logs to a file that is named with the timestamp. Also, generating the JSON values in Scala and then using the `logger.error(...)` command to log them. 

A log so far looks something like this (I still need to add in the file name and path):
```{"errorType":"SchemaValidationError","filePath":"","fileName":"","message":"Data does not conform to schema from https://schema.humancellatlas.org/type/biomaterial/5.1.0/specimen_from_organism; #: required key [schema_type] not found"}```

**An open question:** We need to use Argo + bash (or something else) to write these logs to a Google Bucket. Is there a good way for a bash script to identify the most recent log file? Or maybe it would make more sense to use a static name for the log file and then have the bash script update the name when it copies it to the bucket?

### CODE REVIEW

File: transformation/src/test/scala/org/broadinstitute/monster/hca/HcaPipelineBuilderSpec.scala [hunk 1]
```diff
@@ -288,6 +288,8 @@ class HcaPipelineBuilderSpec
```
snwessel, line 288: @raaidbroad Should the url part of the above `exampleUrlAndFile ` value be a filename instead? It looks like the validateJson method is expecting the name of the metadata file.

snwessel:
I thought about it a bit more and think I answered my own filename question: 

I think it makes sense to have the logger output logs to a static filename (something like `logs/errors.log`). Then a bash script, which can be run by Argo, can check to see if `logs/errors.log` exists and then, if so, it will copy the file to the google bucket and rename it, and then it can delete the local log file (so we start fresh the next time). 

Let me know if anyone has feedback on that plan.

transformation/src/test/scala/org/broadinstitute/monster/hca/HcaPipelineBuilderSpec.scala
raaidbroad: yep you got it

raaidbroad:
@snwessel if it is easy to have the written file from this be errors/{timestamp}.json then we might as well, but if we can do it better in bash, that works fine too. I think your approach is stellar for this type of error; I think we'll be  able to add in a similar thing for errors where the corresponding files don't exist for file metadata (or vice versa) and for general errors on our end where we don't want to provide details but just want to be like "something broke, we're on it", but we might need a little more thought for checksum validation errors and repo API errors since those happen after Dataflow (well I guess we can try to do the checksum stuff in here too). In any case, I'm rambling beyond the scope of this ticket and we'll have to find a way to combine later errors with the errors generated from here anyway; GOOD JOB!

snwessel:
A quick update since the last comments:
- Scala now outputs logs to the file `/logs/errors.log`. Logs are written to the file using the `logger.error(...)` method. An example output of the logger looks like: 
```
{"errorType":"SchemaValidationError","filePath":"/metadata/organoid/123456_VERSION1.json","fileName":"123456_VERSION1.json","message":"Data does not conform to schema from https://schema.humancellatlas.org/type/biomaterial/5.1.0/specimen_from_organism; #: required key [schema_type] not found"}
```
- I added another step to the end of the argo workflow to move the `/logs/errors.log` file to GCS and rename it to the timestamp of the start of the workflow. 
I think this setup should allow us to continue using this setup for other HCA error logging.

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
```diff
@@ -262,9 +265,21 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
   def encode(msg: Msg): String =
     upack.transform(msg, StringRenderer()).toString
 
-  def validateJson(filenamesAndMsg: SCollection[(String, Msg)]): SCollection[(String, Msg)] = {
+  def logSchemaValidationError(filepath: String, errorMessage: String): Unit = {
+    val filenameRegex = "[^/]*$".r // match everything that is not followed by a "/" (only the filename)
+    val filename = filenameRegex.findFirstIn(filepath).getOrElse("")
+    val errorLog = ujson.Obj(
+      "errorType" -> ujson.Str("SchemaValidationError"),
+      "filePath" -> ujson.Str(filepath),
```
snwessel, line 273: This `filepath` value is coming from [here](https://github.com/DataBiosphere/hca-ingest/blob/master/transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala#L109) - which the Apache Beam [documentation](https://beam.apache.org/releases/javadoc/2.1.0/org/apache/beam/sdk/io/fs/ResourceId.html) describes as _the name of the file or directory denoted by this ResourceId. The file name is the farthest element from the root in the directory hierarchy_.

I still can't tell whether it would return the full GCS path or just the part we want, but I don't know of a good way to to test it either than just running the workflow and looking at the logs. Let me know if you have any suggestions.

orchestration/templates/ingest-hca.yaml
```diff
@@ -174,6 +174,10 @@ spec:
                 - name: bq-dataset
                   value: {{ $datasetName | quote }}
 
+          - name: move-error-logs-to-gcs
+            dependencies: [ingest-non-file-metadata, ingest-file-metadata, ingest-data-files]
```
snwessel, line 178: Added these dependencies here because I want this step to run at the very end of the ingest process (so we don't miss any logs).

transformation/src/main/resources/logback.xml
```diff
@@ -5,11 +5,22 @@
...
+
     <!-- Use debug level for our own messages so we can eyeball logs during testing. -->
     <logger name="org.broadinstitute" level="DEBUG" additivity="false">
         <appender-ref ref="STDOUT" />
     </logger>
 
+    <logger name="org.broadinstitute" level="ERROR" additivity="false">
+        <appender-ref ref="FILE" />
+    </logger>
+
```
snwessel, line 23: These changes just make the logger write any error messages to a file. They don't change the way our other logging works.

orchestration/templates/ingest-hca.yaml
```diff
@@ -694,3 +698,22 @@ spec:
...
+      script:
+        image: google/cloud-sdk:slim
+        env:
+          - name: TIMESTAMP
+            value: {{ include "argo.timestamp" . | quote }}
+          - name: BUCKET_NAME
+            value: {{ .Values.gcs.stagingBucketName }}
+        command: [bash]
+        source: |
+        {{- include "argo.render-lines" (.Files.Lines "scripts/move-error-logs-to-gcs.sh") | indent 10 }}
```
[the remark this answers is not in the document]
quazi-broad: I see you changed the name of the file above which makes more sense since it's is doing error specific logging 👍

snwessel:
@quazi-broad sorry to throw in one last change after you approved. It just occurred to me that we would probably want the bash script to fail if it can't find the log file (so that we can see the error in the Argo UI).

I agree with what you said - that watching the Argo UI for failures (and checking that the logs exist in the right place when it's done) might be the best way to test this 😁

orchestration/scripts/move-error-logs-to-gcs.sh
```diff
@@ -0,0 +1,12 @@
...
+declare -r LOGS_DIR=../../logs
+declare -r LOG_FILE=${LOGS_DIR}/errors.log
+declare -r NEW_FILENAME=${LOGS_DIR}/${TIMESTAMP}.log
+
+# rename the file to the start time of the import
+mv ${LOG_FILE} ${NEW_FILENAME}
+# move the file to a bucket
+gsutil cp ${NEW_FILENAME} ${BUCKET_NAME}/errors/
+rm ${NEW_FILENAME}
+
```
raaidbroad, line 12: did you test out if this bash script/series of commands works in general? It looks fine to me, but just want to confirm

orchestration/templates/ingest-hca.yaml
```diff
@@ -694,3 +698,22 @@ spec:
...
         source: |
         {{- include "argo.render-lines" (.Files.Lines "scripts/get-outdated-row-ids.sh") | indent 10 }}
+
+    ##
+    ## Move any logs to GCS.
+    ##
+    - name: move-error-logs-to-gcs
+      inputs:
+        parameters:
+          - name: table
```
raaidbroad, line 708: why are we passing in table here? I don't see it being used, or am I missing it?

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
raaidbroad: I think we should test this before merging it, and unfortunately we don't have a good way set up to do that (I have an idea but it's a longer-term solution). For a quick fix, Sarah, I'll work with you today to coordinate how to test this. We'll make some dummy input with an invalid schema and run this in dataflow.

orchestration/scripts/move-error-logs-to-gcs.sh
snwessel: Yeah, thanks for checking. I tested it using a bucket I had been using to test other things:
```
TIMESTAMP=01-01-2020 BUCKET_NAME=gs://monster-test-bucket/encode-explorer-backfill-requests bash -c './move-error-logs-to-gcs.sh'
```
And it worked (renamed and copied the file to an "errors" folder in the bucket, and then deleted the file locally).

snwessel: This reminded me of one thing that might be worth noting though - I had been running this script from inside the `/orchestration/scripts` directory, but the script doesn't work when you run it from a different directory because it uses a relative path to find the log file. Do you happen to know if there is a way to do this without using a relative path? Or if the relative path I have here would work with the way Argo runs things?

If not, we can see what happens when we run it this afternoon.

orchestration/templates/ingest-hca.yaml
snwessel: Good catch! Thank you! This actually doesn't need any parameters so I just removed it.

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
```diff
@@ -279,34 +295,44 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
           case Some(schema) => {
             // if the schema is not able to load, throw an exception, otherwise try to use it to validate
             Schema.loadFromString(schema) match {
-              case Failure(_) =>
-                throw new Exception(
-                  s"Schema not loaded properly for schema at $url, file $filename"
-                )
+              case Failure(_) => {
+                val errorMessage = s"Schema not loaded properly for schema at $url, file $filename"
+                logSchemaValidationError(filename, errorMessage)
```
pshapiro4broad, line 300: I don't know much if anything about this code at this point, but my usual pattern for code that throws and logs is to log at the catch point, not the throw point. That way you don't have to remember to add a log to every throw. Plus, making sure you log at catch points means that exceptions are always logged.

It looks like in this case you want something special to be logged for this specific exception, in that case throwing a custom object that either itself knows how to log, or can be used to determine a special logging path at the catch point, would let you log this exception differently than others.

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
```diff
@@ -279,34 +295,44 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
...
+                throw new Exception(errorMessage)
```
pshapiro4broad, line 301: The fact that this throws at all is probably not very scala-like. Since the code is operating on a collection, having it return a collection of Either[(String, Msg), ValidationError] would be more what I'd expect, so each returned element would either be the validated tuple or the error for that tuple. But that kind of change depends a lot more on the context in which this is used.

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
```diff
@@ -279,34 +295,44 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
...
+          case None => {
+            val errorMessage = s"No schema found at $url for file $filename"
+            logSchemaValidationError(filename, errorMessage)
+            throw new Exception(errorMessage)
+          }
         }
       }
     }
-    filenamesAndMsg
+    filenameAndMsg
```
pshapiro4broad, line 335: Since this code was modified to not side effect its arguments, it might be clearer to make this method `void` so it's explicit that this method is called purely for its side effects. That would avoid the need to return any value here and you could change
```
    val content = filenameAndMsg.map {
```
to use `for`

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
raaidbroad: This is a very good point; this is my bad for designing it the way I did without an actual catching piece, and this reference material might help with fixing it up https://docs.scala-lang.org/overviews/scala-book/functional-error-handling.html

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
raaidbroad: You're right, it should return something like that, or some type of `Option` or other thing listed here: https://docs.scala-lang.org/overviews/scala-book/functional-error-handling.html

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
raaidbroad: Unfortunately, `SCollection`s don't have a `foreach` method, since they are the collection type used for Beam and Beam strictly doesn't allow side effects

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
raaidbroad: Part of the issue is that each successive step is already returning some kind of `Try` or `Either`; for example, `Schema.loadFromString` returns the type `Try[Schema]`, which needs to be matched on and dug into before moving on to the next step. I'm still upping my functional programming game, but my understanding is that Scala handles Exceptions via pattern matching. I wonder if there is a way to cleanly move all this matching out, and to make sure that the error logging only takes place at a `catch` equivalent.

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
pshapiro4broad: The SCollection and scio stuff is new to me. It's a bit odd but you're right, no foreach..

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
snwessel: Thank you! I agree, that catching it at a single point sounds like the better way to go. I'll go ahead and update it.

snwessel: I updated it to use a try catch but haven't updated the corresponding test yet (it's still expecting an error to be thrown). Maybe we can talk a bit more about what we want the behavior to look like here when the three of us meet this afternoon?

snwessel:
I moved the Argo workflow changes out of this PR into my [checksum validation PR](https://github.com/DataBiosphere/hca-ingest/pull/30) since we'll be refactoring the schema validation soon to not rely at all on Argo.

PR #27, opened 2020-08-14
ra-DSPDC-1247-json-mangle-bug

raaidbroad:
add test to make sure json validation doesn't mess with the json, remove unnecessary coder, make sure validation function returns EXACTLY what was passed to it while it still validates json

### DIFF
--- a/transformation/src/test/scala/org/broadinstitute/monster/hca/HcaPipelineBuilderSpec.scala
+++ b/transformation/src/test/scala/org/broadinstitute/monster/hca/HcaPipelineBuilderSpec.scala
@@ -1,8 +1,6 @@
 package org.broadinstitute.monster.hca
 
-import com.spotify.scio.coders.Coder
 import com.spotify.scio.testing.PipelineSpec
-import io.circe.schema.Schema
 import org.broadinstitute.monster.common.PipelineCoders
 import org.broadinstitute.monster.common.msg.JsonParser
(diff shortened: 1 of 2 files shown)


transformation/src/test/scala/org/broadinstitute/monster/hca/HcaPipelineBuilderSpec.scala
```diff
@@ -290,4 +286,47 @@ class HcaPipelineBuilderSpec
...
+        |            "text": "Homo sapiens",
+        |            "ontology": "NCBITaxon:9606"
+        |        }
+        |    ],
+        |    "describedBy": "https://schema.humancellatlas.org/type/biomaterial/5.1.0/specimen_from_organism"
+        |}
+        |""".stripMargin
+    )
+
+    val exampleUrlAndFile = (exampleFileContent.read[String]("describedBy"), exampleFileContent)
```
snwessel, line 324: Should this be a file path instead of a URL? I think we talked about the same thing [here](https://github.com/DataBiosphere/hca-ingest/pull/25/files#r465936121), but it might have been copied into this test too.

raaidbroad: Good catch, I incorrectly copied and pasted.

PR #28, opened 2020-08-18
Ra dspdc 1237 multiflow

raaidbroad:
Break up the main import workflow into multiple workflows. This way we can run staging and ingestion pieces separately, and we have separate flows for including files or not (which were not in the spec but are useful for testing). The main import workflows are composed of the staging and ingestion workflows, which keeps the code just a little bit dryer than it would've been otherwise.


raaidbroad:
I like the naming suggestions, Sarah; I changed it to "import" because for this project, other folks take issue with us using the term "ingest". I think "load" makes more sense, and I agree with the "table" comment. Will make changes and push.

PR #31, opened 2020-09-18
ra-DSPDC-1284-change-target-path

raaidbroad:
updates targetPath value to contain the filename with extension; will update all getOrElse exception throwing to error logging in a separate PR. Updated test for file ingest request generation as well.


pshapiro4broad (approved):

quazi-broad (approved):
LGTM!

PR #32, opened 2020-09-22
Ra dpsdc 1279 idempotence

raaidbroad:
- Make sure that the file injection script looks at the checksum, not the target path.
- Copy/paste/adjust diffing and appending logic specific to HCA. Ideally we would have generalized the existing code more, but that will take a lot longer than we have, and our future dockerization/Pythonification efforts should address it soon.
- Use new/different diffing and appending logic in existing workflow, make sure we only soft delete old versions, append new rows, and no-op on existing PKs.

### DIFF
--- a/orchestration/scripts/diff-hca-table.sh
+++ b/orchestration/scripts/diff-hca-table.sh
@@ -6,7 +6,6 @@ function join_by { local d=$1; shift; echo -n "$1"; shift; printf "%s" "${@/#/$d
 # Point to BQ metadata we expect to be present on disk.
 declare -r TABLE_DIR=/bq-metadata/${TABLE}
 declare -r PK_COLS=$(cat ${TABLE_DIR}/primary-keys)
-declare -r COMPARE_COLS=$(cat ${TABLE_DIR}/compare-cols)
 
 declare -a DATAREPO_PKS=()
 for col in ${PK_COLS//,/ }; do
--- a/orchestration/scripts/diff-hca-table.sh
+++ b/orchestration/scripts/diff-hca-table.sh
@@ -8,13 +8,6 @@ declare -r TABLE_DIR=/bq-metadata/${TABLE}
 declare -r PK_COLS=$(cat ${TABLE_DIR}/primary-keys)
 declare -r COMPARE_COLS=$(cat ${TABLE_DIR}/compare-cols)
 
-# Build the WHERE clause of the SQL query.
-declare -a COMPARISONS=()
-for pk in ${PK_COLS//,/ }; do
-  COMPARISONS+=("J.${pk} = S.${pk}")
-done
-declare -r FULL_DIFF=$(join_by ' AND ' "${COMPARISONS[@]}")
-
 declare -a DATAREPO_PKS=()
 for col in ${PK_COLS//,/ }; do
   DATAREPO_PKS+=("${col} as datarepo_${col}")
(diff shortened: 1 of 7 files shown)


orchestration/scripts/diff-hca-table.sh
```diff
@@ -0,0 +1,49 @@
+set -euo pipefail
+
+# Copy-pasted from StackOverflow.
+function join_by { local d=$1; shift; echo -n "$1"; shift; printf "%s" "${@/#/$d}"; }
+
+# Point to BQ metadata we expect to be present on disk.
+declare -r TABLE_DIR=/bq-metadata/${TABLE}
+declare -r PK_COLS=$(cat ${TABLE_DIR}/primary-keys)
+declare -r COMPARE_COLS=$(cat ${TABLE_DIR}/compare-cols)
```
pshapiro4broad, line 9: appears to be unused

orchestration/scripts/diff-hca-table.sh
```diff
@@ -0,0 +1,49 @@
+set -euo pipefail
+
+# Copy-pasted from StackOverflow.
+function join_by { local d=$1; shift; echo -n "$1"; shift; printf "%s" "${@/#/$d}"; }
+
+# Point to BQ metadata we expect to be present on disk.
+declare -r TABLE_DIR=/bq-metadata/${TABLE}
```
pshapiro4broad, line 7: In general, all `${VAR}` should be `"${VAR}"` to avoid issues with spaces (although the queries below are OK since they're already in double quotes). In this case I'm guessing TABLE etc don't have spaces but I usually try to do it anyway as a good bash practice.

orchestration/scripts/diff-hca-table.sh
```diff
@@ -0,0 +1,49 @@
...
+declare -r PK_COLS=$(cat ${TABLE_DIR}/primary-keys)
+declare -r COMPARE_COLS=$(cat ${TABLE_DIR}/compare-cols)
+
+declare -a DATAREPO_PKS=()
+for col in ${PK_COLS//,/ }; do
+  DATAREPO_PKS+=("${col} as datarepo_${col}")
+done
+declare -r REPO_KEYS=$(join_by ', ' "${DATAREPO_PKS[@]}")
```
pshapiro4broad, line 15: Note that when you use `declare` on a subshell result you're throwing away the return code of the subshell, and `set -e` won't get triggered if the subshell return code is nonzero. In this case it doesn't matter but it can lead to problems if you're not careful.

For more info see https://github.com/koalaman/shellcheck/wiki/SC2155

pshapiro4broad (approved):
Looks OK as far as I can tell, just a few bash nitpicks.

With all this bash, I can see how it would be nice to have tests.

orchestration/values.yaml
```diff
@@ -14,10 +14,6 @@ argoTemplates:
   exportBQTable:
     create: true
     name: export-table
-  diffBQTable:
```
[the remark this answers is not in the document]
raaidbroad: oh good catch!

quazi-broad (approved):
New YAML files look good. Bash looked good to me, with Phil's comments.
I was trying to see how `helmInjectVersionValues` gets used - why are we removing that here?

raaidbroad:
> New YAML files look good. Bash looked good to me, with Phil's comments.
> I was trying to see how `helmInjectVersionValues` gets used - why are we removing that here?

It injects an Argo template to reference, but we no longer reference it since I had to make a new one for HCA's update logic.

PR #33, opened 2020-09-23
Ra dspdc 1238 file error log

raaidbroad:
Will merge the idempotence bug fix before this. This adds warning and error logs for a number of things, all enumerated in the HcaLog file. It also uses the new hook in for a post process function for our pipelines, so if even a single error is logged, we fail the pipeline.


transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipeline.scala
```diff
@@ -1,8 +1,11 @@
 package org.broadinstitute.monster.hca
 
+import caseapp.core.help.Help
+import caseapp.core.parser.Parser
 import org.broadinstitute.monster.common.{PipelineBuilder, ScioApp}
 
 /** Entry-point for the HCA pipeline's Docker image. */
-object HcaPipeline extends ScioApp[Args] {
+object HcaPipeline
+    extends ScioApp[Args]()(Parser[Args], Help[Args], postProcess = PostProcess.postProcess) {
```
pshapiro4broad, line 9: Hmm, maybe a more natural way to express this would've been to make `postProcess` an overrideable `def` in `ScioApp`, just like `pipelineBuilder` is. Making it a class constructor arg seems a bit awkward and the value isn't needed at class construction time.

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaLog.scala
```diff
@@ -0,0 +1,60 @@
...
+case class HcaErrorLog() extends LogLevel
+case class HcaWarnLog() extends LogLevel
+
+// a trait to capture the generic logging mechanism we'll want, with case classes for the different logging levels
+abstract class HcaLog {
+  val level: LogLevel
+  val jsonMsg: Obj
+
+  def log(implicit logger: Logger): Unit =
+    level match {
```
pshapiro4broad, line 18: I don't think `level` is necessary. If the `log` method is `abstract` here then you can override it in `HcaErrorLog` and `HcaWarnLog` to do the correct thing for each case.

pshapiro4broad (commented):
Just a few comments so far, but wanted to push them right away in case you wanted to act on them.

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaLog.scala
```diff
@@ -0,0 +1,60 @@
...
+    .Obj(
+      "errorType" -> ujson.Str(this.getClass.getSimpleName),
+      "filePath" -> ujson.Str(filepath),
+      "fileName" -> ujson.Str(filename),
+      "message" -> ujson.Str(msg)
+    )
+}
+
+// case classes for all the different actual warnings and errors we want to raise during the workflow
+case class NoMatchWarning(msg: String) extends HcaWarn(msg)
```
quazi-broad, line 52: Have we hit instances of each of these types of warnings/errors?

quazi-broad (approved):
Warning/error classes were easy to follow and understand. Changes look good to me.

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipeline.scala
raaidbroad: Can do; I think that's definitely cleaner. I'll make that change and update to a newer version of that library then make the change here.

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaLog.scala
raaidbroad: Ah good call, I like that approach better.

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaLog.scala
raaidbroad: I added warnings and errors for things we've been encountering and things we agreed to in the MVP spec.

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
```diff
@@ -85,20 +107,35 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
...
         FileIO
           .`match`()
-          .filepattern(jsonPath)
+          .filepattern(filePattern)
           .withEmptyMatchTreatment(EmptyMatchTreatment.ALLOW_IF_WILDCARD)
       )
-    }.applyTransform[ReadableFile](FileIO.readMatches())
+    }
+    metadata.count
+      .map(c => if (c == 0.toLong) NoMatchWarning(s"$filePattern had $c matches").log)
```
pshapiro4broad, line 125: I don't think `.toLong` is needed here.

Will this work? `map` returns a new `SCollection` but nothing reads it, so would it get forced to execute?

raaidbroad: It was executed when I was playing around with it 🙂

raaidbroad:
Test bucket for validation tests is here: https://console.cloud.google.com/storage/browser/broad-dsp-monster-hca-dev-test-storage;tab=objects?forceOnBucketsSortingFiltering=false&project=broad-dsp-monster-hca-dev&prefix=&forceOnObjectsSortingFiltering=false

pshapiro4broad (commented):
I'm still looking but am pushing my comments so far. More suggestions but maybe it's time to move on...

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
```diff
@@ -151,38 +180,48 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
...
+    inputPrefix: String
+  ): Option[Msg] =
+    getEntityIdAndVersion(fileName, inputPrefix, entityType).map {
+      case (entityId, entityVersion) =>
+        Obj(
+          Str(s"${entityType}_id") -> Str(entityId),
+          Str("version") -> Str(entityVersion),
+          Str("content") -> Str(encode(metadata)),
+          Str("crc32c") -> descriptor.tryRead[Msg]("crc32c").getOrElse {
+            MissingPropertyError(
```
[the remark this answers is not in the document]
raaidbroad: Yeah this was bothering me too; ooo the filter idea is good! I def want it to filter this out upstream because yeah we are hitting the same thing twice. I might as well just do it now

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
```diff
@@ -429,20 +419,110 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
...
+                MissingPropertyError(
+                  s"$inputPrefix/$filename",
+                  "File has no describedBy property."
+                ).log
+                None
+              }(Some(_)),
+            (filename, msg)
+          )
+      }
+      .collect { case (Some(url), rest) => (url, rest) }
```
pshapiro4broad, line 471: There must be a way to express this using `filter` but I couldn't get it it work. The code is doing `map(.. function that creates optional elements)` and then `collect(... function that removes optional elements)`. So in theory you should be able to combine both into a `filter`..

pshapiro4broad (approved):
LGTM; there will always be more things to improve but I think this has the correct functionality. It would be nice to get feedback from a scala guru since it seems like there should be a better way of doing a lot of this.

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
raaidbroad: I think we probably want to do a similar thing as your suggestion for the crc32c; I can filter and log upstream before it even gets here

pshapiro4broad (approved):
Looks even better!

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
pshapiro4broad: Yeah, and I realized that it did need both map and collect (or map and filter) so there wasn't a better way to do it here.

PR #36, opened 2020-10-13
Ra dspdc 1309 graph too large

raaidbroad:
Fix graph-too-large error and add a Dataflow Runner check to make sure the pipeline works on Dataflow.

### DIFF
--- a/transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
+++ b/transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
@@ -26,12 +26,6 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
     // Is the count of metadata/{file_type} == descriptors/{file_type}? if not, raise warns
     // Is the count of data/** == metadata/** == descriptors/**? if not, raise warns
 
-    // need to include the input prefix here so that a join can work later for validation
-    val dataFiles =
-      matchFiles(s"${args.inputPrefix}/data/**", ctx).keyBy(_.resourceId().getFilename).map {
-        case (filename, rest) => (s"${args.inputPrefix}/data/$filename", rest)
-      }
-
     val allMetadataEntities =
       expectedMetadataEntities.map(_ -> false) ++ expectedFileEntities.map(_ -> true)
     allMetadataEntities.foreach {
@@ -41,7 +35,6 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
           args.inputPrefix,
           args.outputPrefix,
           entityType,
-          dataFiles,
           isFileMetadata
         )
     }
@@ -291,25 +284,6 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
   def encode(msg: Msg): String =
     upack.transform(msg, StringRenderer()).toString
 
-//  def filterProperty(
-//    messages: SCollection[(String, Msg)],
-//    inputPrefix: String,
-//    property: String
-//  ): SCollection[(String, Msg)] = {
-//    messages.withName(s"Filtering files in $inputPrefix for property: $property").filter {
-//      case (filename, message) =>
-//        message
-//          .tryRead[String](property)
-//          .fold {
-//            MissingPropertyError(
-//              s"$inputPrefix/$filename",
-//              s"File has no $property property."
-//            ).log
-//            false
-//          }(_ => true)
-//    }
-//  }
-
   /**
     * @param filename the filename to use if the property is missing
     * @param msg the message body to check the property for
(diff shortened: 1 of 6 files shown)


transformation/src/main/scala/org/broadinstitute/monster/hca/PostProcess.scala
```diff
@@ -1,13 +1,15 @@
...
+import com.spotify.scio.ScioResult
 
 object PostProcess {
-  val errorCount: Counter = ScioMetrics.counter("errorCount")
 
   def postProcess(result: ScioResult): Unit = {
-    result.counter(errorCount).committed.fold(())(count => if (count > 0) throw new HcaFailException)
+    result.allCounters.foreach {
+      case (name, count) =>
+        if (name.getName == "errorCount")
```
pshapiro4broad, line 10: Maybe put `"errorCount"` into a declared constant to make it clear that it's the same thing everywhere this string is repeated.

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
```diff
@@ -543,6 +505,8 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
           case None =>
             Option(SchemaValidationError(errPath, s"No schema found at $url"))
         }
+        a.foreach { err => err.log }
```
pshapiro4broad, line 508: Since you don't need `a` after this. you could write it without it. Also I think you can use an anonymous variable here:
```
        schemaOption match {
            ...
        }.foreach(_.log)
```

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
```diff
@@ -460,35 +440,17 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
...
-    filenamesAndMsg
-  }
-
   /**
     * @param inputPrefix The entire prefix before the filename
     * @param filenamesAndMsg The filenames and messages to validate
     * @return A collection of possible errors
     */
-  def validateJsonInternal(inputPrefix: String)(
+  def validateJson(inputPrefix: String)(
```
pshapiro4broad, line 448: I don't see why you need two arg lists for this function, does that affect the graph generation?

pshapiro4broad (approved):
Looks good.

transformation/src/main/scala/org/broadinstitute/monster/hca/PostProcess.scala
raaidbroad: Good call. I was annoyed to have to do it this way, apparently when we use the dataflow runner the metric has to be declared inside a block of code that is run in the pipeilne (that restriction doesn't apply to the direct runner), but a constant for the string is good.

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
raaidbroad: The only reason was for running the function with input data in the tests; didn't have a good way of running it otherwise :/

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
```diff
@@ -392,21 +384,10 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
...
       // Generate file ingest requests from descriptors. Deduplicate by the content hash.
-      val keyedIngestRequests = validatedDescriptors.map {
+      val fileIngestRequests = validatedDescriptors.map {
         case (_, descriptor) =>
           generateFileIngestRequest(descriptor, inputPrefix)
-      }.distinctByKey.values.keyBy(request => request.read[String]("source_path"))
-
-      val fileIngestRequests =
-        keyedIngestRequests.leftOuterJoin(files).flatMap {
-          case (filename, (_, None)) =>
```
quazi-broad, line 402: Is this the extra validation that you mentioned "might cause issues in the future"?

raaidbroad: yeah, because there is a future where we want to update metadata about files that are already in the repo

quazi-broad (approved):
Changes look good to me - as I understand it you removed some extra validation in order to trim some fat from the graph file output? Is there any way for a user to see how big that file becomes with a specific run or do you only see that piece when the file is too big?

raaidbroad:
> Changes look good to me - as I understand it you removed some extra validation in order to trim some fat from the graph file output? Is there any way for a user to see how big that file becomes with a specific run or do you only see that piece when the file is too big?

We can see the output job file (which is not the same as the output graph) and look at how many steps there are, as well as a bunch of details about each step. Keeping the number of steps lower reduces the graph size. The error only tells us that it is too big, and we can see the output job file by adding a --output-job-file or some similar flag when running the pipeline.

PR #38, opened 2020-11-23
Ra no max failures

raaidbroad:
## Why
Bulk file ingests should fail fast particularly when there is the same issue for all the files and all of them fail.
Dataflow and Argo could both perform a little better with more resources, which is helpful for getting imports done quickly.

## This PR
- Set max failures for bulk file ingests to 0 so that Jade will fail fast on file ingests
- Bump up dataflow machine type and max number of worker params
- Bump up parallelism for argo workflows

### DIFF
--- a/orchestration/values.yaml
+++ b/orchestration/values.yaml
@@ -1,9 +1,9 @@
 dataflow:
   region: us-central1
-  workerMachineType: n1-standard-2
+  workerMachineType: n1-standard-4
   autoscaling:
     minWorkers: 4
-    maxWorkers: 8
+    maxWorkers: 16
   useFlexRS: false
 repo:
   pollTimeout: 86400
@@ -20,4 +20,4 @@ argoTemplates:
   softDeleteTable:
     create: true
     name: soft-delete-table
-parallelism: 4
+parallelism: 8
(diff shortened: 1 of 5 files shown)


aherbst-broad:
`when there is the same issue` --> can you clarify: operationally you're observing that ingests are dying consistently with the same error on all files and we're wasting time attempting to ingest the rest of the data set, hence this PR?

raaidbroad:
> `when there is the same issue` --> can you clarify: operationally you're observing that ingests are dying consistently with the same error on all files and we're wasting time attempting to ingest the rest of the data set, hence this PR?

Yes

orchestration/values.yaml
```diff
@@ -20,4 +20,4 @@ argoTemplates:
   softDeleteTable:
     create: true
     name: soft-delete-table
-parallelism: 4
+parallelism: 8
```
aherbst-broad, line 23: ❓ I am curious; Looking at https://cloud.google.com/compute/docs/machine-types I see that an `n1-standard-4` has 4 hyper-threads, why the bump to parallelism 8 here given that capacity constraint?

raaidbroad: This parallelism is for Argo, it refers to the number of pods a single workflow is allowed to spin up at a time.

aherbst-broad (approved):
Seems reasonable; (quick nit: consider using a more descriptive string for the PR name)

orchestration/values.yaml
raaidbroad: An example of it being used downstream: https://github.com/DataBiosphere/hca-ingest/blob/7b565ae53c37feb802a9166109a3c8c8fc4fce33/orchestration/templates/import-hca-total.yaml#L26

quazi-broad (approved):
Changes lgtm

PR #40, opened 2020-12-11
Add hca-mvp "runner" workflows

aherbst-broad:
## Why
We currently kick off our hca workflows by cutting and pasting them out of a shared google doc and into the argo web UI. We should get those in source control and kick them off via the argo CLI tool.

## This PR
* Creates the `run-import-hca-total` and `run-stage-data` workflows. These can be invoked via the `argo` tool in the terminal.


raaidbroad (commented):
❓ Do you think it makes sense to include an example of running each of these with the Argo CLI, perhaps by adding to a readme?

ThePletch (approved):
still getting my head around argo, but this looks like it'll work

raaidbroad (approved):

PR #41, opened 2020-12-11
Ra dspdc 1368 duplicates script

raaidbroad:
## Why
We have had more than one snapshot handed off to the indexer where there are duplicate rows (as in the same `{entity}_id` has multiple entries in a table) or there are null file references. This really, really slows down iteration because snapshots can take a long time. We therefore have strong incentive to have a simple way to check and fix any issues with a dataset before creating a snapshot from it. Ideally this would reduce the number of iterations and not push back deadlines more than necessary.

## This PR
Adds a simple Python CLI to check and fix HCA datasets that have duplicate rows or null file references. It does so using the Poetry package management tool, which will likely require installation, and also depends on some manual google auth setup.


scripts/hca-utils/hca_utils/utils.py
```diff
@@ -0,0 +1,332 @@
...
+        jade_urls = {"prod": "https://jade-terra.datarepo-prod.broadinstitute.org",
+                     "dev": "https://jade.datarepo-dev.broadinstitute.org"}
+        self.base_url = jade_urls[environment]
+
+        # this is always prod for bigquery, storage, etc.
+        gcp_creds, _ = google.auth.load_credentials_from_file("path/to/prod/credentials.json")
+        self.gcp_creds = gcp_creds
+
+        # this depends on the Jade env to interact with
+        creds_path = {"prod": "path/to/prod/credentials.json",
```
raaidbroad, line 43: I am not proud of this

scripts/hca-utils/hca_utils/utils.py
```diff
@@ -0,0 +1,332 @@
...
+                           "dev": "broad-dsp-monster-hca-dev"}
+        self.bucket_project = bucket_projects[environment]
+        self.bucket = f"broad-dsp-monster-hca-{environment}-staging-storage"
+
+        jade_urls = {"prod": "https://jade-terra.datarepo-prod.broadinstitute.org",
+                     "dev": "https://jade.datarepo-dev.broadinstitute.org"}
+        self.base_url = jade_urls[environment]
+
+        # this is always prod for bigquery, storage, etc.
+        gcp_creds, _ = google.auth.load_credentials_from_file("path/to/prod/credentials.json")
```
raaidbroad, line 39: I am not proud of this

scripts/hca-utils/hca_utils/utils.py
```diff
@@ -0,0 +1,332 @@
...
+        """
+        query = f"""
+        WITH fileRefTables AS (SELECT * FROM `{self.project}.datarepo_{self.dataset}.INFORMATION_SCHEMA.COLUMNS` WHERE column_name = "file_id"),
+        desiredViews AS (SELECT * FROM `{self.project}.datarepo_{self.dataset}.INFORMATION_SCHEMA.TABLES` WHERE table_type = "VIEW")
+        SELECT desiredViews.table_name FROM fileRefTables JOIN desiredViews ON fileRefTables.table_name = desiredViews.table_name
+        """
+
+        return self._hit_bigquery(query)
+
+    def get_duplicates(self, target_table: str) -> Set[str]:
```
raaidbroad, line 74: The query in this function is A LOT so please ask questions; I tried to use clear names and commented with my thought process

scripts/hca-utils/hca_utils/utils.py
```diff
@@ -0,0 +1,332 @@
...
+
+    def get_null_filerefs(self, target_table: str) -> Set[str]:
+        """
+        Determines what rows have null values in the file_id column. We want to soft delete those.
+        :param target_table: The particular table to operate on.
+        :return: A set of row ids to soft delete.
+        """
+        query = f"""
+        SELECT datarepo_row_id
+        FROM `{self.project}.datarepo_{self.dataset}.{target_table}` WHERE file_id IS NULL
```
raaidbroad, line 114: I usually do SQL things in Python with Postgres and therefore psycopg2 which has modules to help be safer about writing your SQL strings; if there are any suggestions on that front, please help! I don't like writing this stuff out like this where we could very easily muck things up. I guess there's no users besides us so 🤷🏾

scripts/hca-utils/hca_utils/utils.py
```diff
@@ -0,0 +1,332 @@
...
+    def remove_all(self):
+        """
+        Check and print the number of duplicates and null file references for each table in the dataset, then soft
+        delete the problematic rows.
+        :return:
+        """
+        self.remove_duplicates()
+        self.remove_null_file_refs()
+
+    def _process_rows(self, get_table_names, get_rids, soft_delete: bool = True):
```
raaidbroad, line 312: This runs synchronously; if we care enough, I can try to make it async, but I figured it isn't really worth it for this

scripts/hca-utils/hca_utils/main.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def run(arguments=None):
+    parser = DefaultHelpParser(description='A simple HcaUtils CLI.')
+    parser.add_argument('-V', '--version', action='version', version='%(prog)s ' + hca_utils_version)
+    parser.add_argument('-e', '--env', help='The Jade environment to target, defaults to dev', choices=['dev', 'prod'],
+                        default='dev')
+    parser.add_argument('-p', '--project', help="The Jade project to target, defaults to correct project for dev")
+    parser.add_argument('-d', '--dataset', help="The Jade dataset to target")
+    parser.add_argument('-t', '--task', help="Just check for problems or perform a soft delete to address the problems",
```
ThePletch, line 23: semantically i usually prefer 'do a thing' vs 'just check to see what needs done' to be defined by the presence/absence of a flag, rather than choosing between two mandatory values.

scripts/hca-utils/hca_utils/utils.py
```diff
@@ -0,0 +1,332 @@
...
+from google.cloud import bigquery, storage
+from requests_cache.core import CachedSession
+
+
+class HcaUtils:
+    def __init__(self, environment: str, project: str, dataset: str):
+        self.environment = environment
+
+        if environment == "dev":
+            self.project = "broad-jade-dev-data"
```
ThePletch, line 19: if we force-overwrite the value of `project` if the env is dev, i think that `project` should be an optional param in the initializer here and be rejected by the argument parser if it's passed in when the env is dev. that'll better surface what's happening to users of the script

scripts/hca-utils/hca_utils/utils.py
ThePletch: if you import `from google.oauth2.service_account import ServiceAccount` you can do this instead:
```suggestion
        self.gcp_creds = service_account.Credentials.from_service_account_file("path/to/prod/credentials.json")
```

scripts/hca-utils/hca_utils/utils.py
ThePletch: i think it might help readability/extensibility to pull the shared snippets out into helper functions, or even build out an entirely separate `BigqueryQueryBuilder` class. that way you'd be able to use descriptive function names to make these verbose comments less necessary.

scripts/hca-utils/hca_utils/utils.py
raaidbroad: maybe a simple thing could be to just have environment variables point to these paths? Y'all would still need to go through the motions of having separate creds files and pointing to them though...

scripts/hca-utils/hca_utils/utils.py
ThePletch: it's definitely worth diving into the bigquery library you're using to see if it supports parameterization. happy to take a look with you, i have a lot of experience jamming my hand into the source code of poorly documented python APIs.

scripts/hca-utils/hca_utils/utils.py
```diff
@@ -0,0 +1,332 @@
...
+        FROM `{self.project}.datarepo_{self.dataset}.{target_table}` WHERE file_id IS NULL
+        """
+
+        return self._hit_bigquery(query)
+
+    def _hit_bigquery(self, query):
+        """
+        Helper function to consistently interact with biqquery while reusing the same client.
+        :param query: The SQL query to run.
+        :return: A set of whatever the query is asking for (assumes that we're only asking for a single column).
```
ThePletch, line 123: might be worth seeing if bigquery returns column names as well, so you can build out dict-based results instead.

scripts/hca-utils/hca_utils/utils.py
```diff
@@ -0,0 +1,332 @@
...
+
+    # local csv interactions
+    def create_csv(self, row_ids: Set[str], target_table: str) -> str:
+        """
+        Create a csv locally with one column filled with row ids to soft delete.
+        :param row_ids: A set of row ids to soft delete.
+        :param target_table: The table that the row ids belong to.
+        :return: The filename of the created csv.
+        """
+        filename = self.filename_template.format(table=target_table)
```
ThePletch, line 136: consider using a function rather than calling `format()` on a fixed string

scripts/hca-utils/hca_utils/utils.py
```diff
@@ -0,0 +1,332 @@
...
+        with open(filename, mode="w") as soft_delete_file:
+            sd_writer = csv.writer(soft_delete_file, delimiter=",", quotechar='"', quoting=csv.QUOTE_MINIMAL)
+
+            for rid in row_ids:
+                sd_writer.writerow([rid])
```
ThePletch, line 141:
```suggestion
            sd_writer.writerows([[rid] for rid in row_ids])
```

scripts/hca-utils/hca_utils/utils.py
```diff
@@ -0,0 +1,332 @@
...
+        return filename
+
+    @staticmethod
+    def delete_csv(filename: str):
+        """
+        Remove the csv after it has been uploaded so that there isn't a deluge of csv files in the local directory.
+        :param filename: The filename of the file to delete.
+        :return:
+        """
+        os.remove(filename)
```
ThePletch, line 152: consider using a temporary file in a context manager (e.g. `with TemporaryFile('filename.csv') as tempfile:`) instead of manually remembering to clean up after yourself

scripts/hca-utils/hca_utils/utils.py
```diff
@@ -0,0 +1,332 @@
...
+    # jade interactions
+    @staticmethod
+    @lru_cache()
+    def _get_endpoint(endpoint: str):
+        """
+        Get relevant information, such as the HTTP method, the path, and parameters, for Jade API endpoints.
+        :param endpoint: The endpoint name to get information for.
+        :return: A dictionary containing the endpoint's method, path, and parameters.
+        """
+        url = "https://jade-terra.datarepo-prod.broadinstitute.org/v2/api-docs"
```
ThePletch, line 180: let's make this a constant at a class level instead

scripts/hca-utils/hca_utils/utils.py
```diff
@@ -0,0 +1,332 @@
...
+
+        my_map = {}
+
+        for path, methods in paths.items():
+            for method, rest in methods.items():
+                if "parameters" in rest:
+                    params = [(param["name"], param["in"]) for param in rest["parameters"]]
+                else:
+                    params = []
+                my_map.update({rest["operationId"]: {"method": method.upper(), "path": path, "params": params}})
```
ThePletch, line 192: this is a triply-nested for loop - let's see if we can break loop iterations into helper functions to make it a little easier to parse.

scripts/hca-utils/hca_utils/utils.py
```diff
@@ -0,0 +1,332 @@
...
+        if endpoint in my_map:
+            return my_map[endpoint]
+        else:
+            raise KeyError(f"Endpoint named {endpoint} not found!")
```
ThePletch, line 196:
```suggestion
        try:
            return my_map[endpoint]
        except KeyError:
            raise KeyError(f"Endpoint named {endpoint} not found!")
```

weirdly, python best practice is to ask forgiveness, not permission

scripts/hca-utils/hca_utils/utils.py
```diff
@@ -0,0 +1,332 @@
...
+        :param query: Optional query parameters for the API request if needed.
+        :return: The result of running handle_ok on the server's response, or an HTTPError.
+        """
+        ep_info = self._get_endpoint(endpoint)
+        url = f"{self.base_url}{ep_info['path']}"
+
+        if params:
+            url = url.format(**params)
+        if query:
+            url = f"{url}/?{urllib.parse.urlencode(query)}"
```
ThePletch, line 214: i see a lot of potential footguns here if someone, e.g., puts a forward slash at the end of the `url` argument or mixes up `params` and `query`. could we use python's built-in `URL` class to help us generate the URL here?

scripts/hca-utils/hca_utils/utils.py
```diff
@@ -0,0 +1,332 @@
...
+
+        response = AuthorizedSession(self.jade_creds).request(method=ep_info["method"], url=url, json=body)
+
+        if response.ok:
+            return handle_ok(response)
```
ThePletch, line 219: why not just return the response in this case rather than require the user to pass in a `handle_ok` callback?

scripts/hca-utils/hca_utils/utils.py
```diff
@@ -0,0 +1,332 @@
...
+        """
+        Get the dataset ID of the provided dataset name.
+        :return: The dataset id.
+        """
+
+        def handle_response(response):
+            return response.json()["items"][0]["id"]
+
+        dataset_id = self._hit_jade("enumerateDatasets", handle_response, query={"filter": self.dataset})
+        return dataset_id
```
ThePletch, line 233:
```suggestion
        jade_response = self._hit_jade("enumerateDatasets", query={"filter": self.dataset})
        return jade_response.json()["items"][0]["id"]
```
example shortened method if we ditch the `handle_ok` callback

scripts/hca-utils/hca_utils/utils.py
```diff
@@ -0,0 +1,332 @@
...
+        """
+        Helper function to consistently interact with biqquery while reusing the same client.
+        :param query: The SQL query to run.
+        :return: A set of whatever the query is asking for (assumes that we're only asking for a single column).
+        """
+        query_job = self.bigquery_client.query(query)
+        return {row[0] for row in query_job}
+
+    # local csv interactions
+    def create_csv(self, row_ids: Set[str], target_table: str) -> str:
```
ThePletch, line 129: since this is specifically making a row id csv, let's include that in the name of this method

scripts/hca-utils/hca_utils/utils.py
ThePletch: (also, i'm thinking we can get away with replacing invocations of this function with just calling `os.remove` directly if we want to stay with this pattern)

scripts/hca-utils/hca_utils/utils.py
```diff
@@ -0,0 +1,332 @@
...
+        """
+        self._process_rows(self.get_all_table_names, self.get_duplicates)
+
+    def remove_null_file_refs(self):
+        """
+        Check and print the number of null file references for each table in the dataset, then soft delete the
+        problematic rows.
+        :return:
+        """
+        self._process_rows(self.get_file_table_names, self.get_null_filerefs)
```
ThePletch, line 301: i'm thinking this should be combined with the `check_for_null_file_refs` method, with the soft delete flag provided as an argument to the combined method. as-is, this trips my "code duplication" spidey sense

scripts/hca-utils/hca_utils/utils.py
ThePletch: worth making this async if it takes eternity to run and is mostly waiting for external services, but otherwise async is currently extremely fiddly in python

ThePletch (commented):
this looks great! thanks for writing this. let's think about reorganizing things a bit to reduce repetition and aid in readability.

overall, `HcaUtils` also strikes me as a bit of a bag of functions. i'm wondering if we can identify different categories of function in it and split them out into their own classes to try and streamline them a bit more and keep more data at the instance level rather than being passed around as arguments/constants.

scripts/hca-utils/hca_utils/utils.py
ThePletch: i think this comment ended up on the wrong line - this is meant to refer to the `remove_null_file_refs` and `check_for_null_file_refs` methods

scripts/hca-utils/hca_utils/utils.py
ThePletch: the best solution here is to read the creds from vault IMO

scripts/hca-utils/hca_utils/utils.py
```diff
@@ -0,0 +1,332 @@
...
+        blob.upload_from_filename(filename)
+
+        filepath = f"gs://{self.bucket}/{filename}"
+        print(f"Put a soft-delete file here: {filepath}")
+
+        return filepath
+
+    # jade interactions
+    @staticmethod
+    @lru_cache()
```
aherbst-broad, line 173: ❓ Is this caching needed, i.e., what are the performance issues this is addressing?

scripts/hca-utils/hca_utils/utils.py
aherbst-broad: fwiw, I think google.auth already [supports](https://google-auth.readthedocs.io/en/latest/reference/google.auth.html?highlight=load_credentials_from_file#google.auth.load_credentials_from_file) an env var for this purpose (GOOGLE_APPLICATION_CREDENTIALS).

I'd be fine with an env var, I have no opinion about vault for this purpose. If you're going to stick with this approach, can the path be config'ed out or passed in via a cli arg?

scripts/hca-utils/hca_utils/utils.py
aherbst-broad: I'm 👎 on async until it's absolutely necessary from a perf perspective, especially in a utility script like this.

scripts/hca-utils/hca_utils/utils.py
aherbst-broad: +1, I think it'd help focus things to extract the db concerns into a separate class.

scripts/hca-utils/hca_utils/utils.py
aherbst-broad: It appears as though BQ and this lib support [parameterization](https://cloud.google.com/bigquery/docs/parameterized-queries), but not for table names.

scripts/hca-utils/hca_utils/utils.py
aherbst-broad: +1, A bit more color can be found in the python [docs](https://docs.python.org/3/glossary.html#term-eafp)

scripts/hca-utils/hca_utils/utils.py
raaidbroad: Since this is for a user (aka one of us) running it, I'm open to storing creds in vault but I would think that's better for a service account (at the very least, Monster hasn't stored personal creds in vault, but if we can/should then sure!)

Yup, that is the env var, but the issue is that we have more than one credentials file to use here. If we set the env var or let it use the default location of the file, and you have run `gcloud auth application-default login` and are logged in with your Broad account, then it'll work great for prod. But if you want to use the dev repo and you have a separate gmail account for that, then this doesn't work because:

- you need your Broad account's credentials for interacting with GCP
- you need your dummy dev account's credentials for interacting with the dev Repo

I suppose another solution here is to simply add our dev email accounts to our dev GCP projects maybe? The dev accounts can already access BigQuery because of Jade ACL setting stuff.

Thoughts @aherbst-broad @ThePletch ?

scripts/hca-utils/hca_utils/utils.py
raaidbroad: Yeah this all runs fast enough that I don't think we should bother

scripts/hca-utils/hca_utils/utils.py
raaidbroad: Sure Any particular reason why?

scripts/hca-utils/hca_utils/utils.py
aherbst-broad, line 180: ❓ Do we need to be pulling the endpoint metadata dynamically like this? I think we're using 2 (`applyDatasetDataDeletion` and `enumerateDatasets`), would it be simpler to just hardcode those for now?

scripts/hca-utils/hca_utils/utils.py
raaidbroad: HAH yep that would make more sense. I've been living in Scala land too long and try to pass around functions at every possible opportunity 😂

aherbst-broad (commented):
Looks good so far!

scripts/hca-utils/hca_utils/utils.py
ThePletch: abstraction, mainly - it's a bit clearer at the point of invocation if you call `self.row_id_filename_for_table(target_table)` rather than `self.filename_template.format(table=target_table)`, but it also lets you freely modify how you generate that filename later

scripts/hca-utils/hca_utils/utils.py
raaidbroad: Probably not needed, just figured that since the API structure doesn't change very often we don't want to create this map for every function invocation

scripts/hca-utils/hca_utils/utils.py
raaidbroad: Probably simpler but I got carried away here and wanted a generic solution in case we were going to add the polling for retrieveJob (to poll until completion) and if we'd want to extend this to also hit the snapshot creation endpoint. Ideally we'd just use the Jade python client but at the time of writing this, I had not yet been provided with the Python client 🤷🏾 I can hard code if desired

scripts/hca-utils/hca_utils/utils.py
ThePletch: hm. given that this is a task that we'll be running as part of production work, my instinct is that we shouldn't be running it with personal credentials in the first place (generally i think any work that's run on live data should be run with a SA that has the permissions it needs and no more, for both auditing purposes and purposes of self-protection, to make sure the script simply isn't allowed to do anything it's not intended to do).

open to discussion if people think that's overkill for this script or otherwise isn't a good idea.

aherbst-broad: I disagree re: a shared service account:
1. What happens when someone leaves the org?
2. We get _less_ auditabilty with a shared acct since all ops will be attributed to the shared account, but potentially run by all devs

Either way, that doesn't help the nut of the problem (needing a dev acct for jade-dev and a prod account for gcp). How hard would be it be to add the dev accts to gcp? I think I like separating the prod from the dev credentials to minimize risk

scripts/hca-utils/hca_utils/utils.py
aherbst-broad: 🤔 well, this is done and it works, I'm ok keeping it.

scripts/hca-utils/hca_utils/utils.py
aherbst-broad: I suggest removing it then; caching unless it's absolutely necessary is a sure route to hard to diagnose bugs.

scripts/hca-utils/hca_utils/utils.py
raaidbroad: Okay so if we want to just address the issue of adding dev emails to dev GCP projects and then we don't have to do any of this file stuff (it'd just be your standard google application default credentials), we could create a user group for monster-dev emails and add them to our dev projects? Idk if that breaks any secops guidance

quazi-broad (approved):
I've looked at this PR a few times but I was mostly just following the comments - in general it looks good to me and I think I understand the different questions brought up so I am approving this PR.

scripts/hca-utils/hca_utils/main.py
```diff
@@ -0,0 +1,44 @@
...
+def run(arguments=None):
+    parser = DefaultHelpParser(description="A simple HcaUtils CLI.")
+    parser.add_argument("-V", "--version", action="version", version="%(prog)s " + hca_utils_version)
+    parser.add_argument("-e", "--env", help="The Jade environment to target, defaults to dev", choices=["dev", "prod"],
+                        required=True)
+    # only allow if env is prod
+    parser.add_argument("-p", "--project", help="The Jade project to target, defaults to correct project for dev")
+    parser.add_argument("-d", "--dataset", help="The Jade dataset to target", required=True)
+    parser.add_argument("-r", "--remove",
+                        help="Remove problematic rows. If flag not set, will only check for presence of problematic rows.",
```
aherbst-broad, line 25: ❤️ I love this "dry run by default" mode of operation.

aherbst-broad (approved):
LGTM, I still think it'd be worth factoring the db concerns into a separate class but this is good enough for now.

PR #42, opened 2021-01-12
ra DSPDC 1382 fix join

raaidbroad:
## Why
We've had a number of issues due to crc32c checksums not being unique across the number of files we have to deal with. When injecting file ids, we need to check more than just the crc32c. Per import, checking by the source file path should be unique.

## This PR
make file-id-injection join use a combo of the checksum and the source file name; the source file name for a given import must be unique, so no collisions should occur


ThePletch (approved):

quazi-broad (approved):
lgtm

PR #43, opened 2021-01-13
Ra dspdc 1348 fix remove script

raaidbroad:
## Why
The "remove" option for the hca post processing tool was no longer working as a result of some refactors having to do with temporary files.

## This PR
Updates the code to make sure the file creation, writing, reading/uploading, and deletion work as intended.

### DIFF
--- a/scripts/hca-utils/hca_utils/utils.py
+++ b/scripts/hca-utils/hca_utils/utils.py
@@ -144,7 +144,7 @@ class HcaUtils:
         bucket = storage_client.bucket(self.bucket)
         target_filename = self._format_filename(table=target_table)
         blob = bucket.blob(target_filename)
-        blob.upload_from_file(local_file, rewind=True)
+        blob.upload_from_file(local_file)
 
         filepath = f"gs://{self.bucket}/{target_filename}"
         print(f"Put a soft-delete file here: {filepath}")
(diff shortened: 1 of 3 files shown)


quazi-broad (approved):
changes look good to me

scripts/hca-utils/hca_utils/utils.py
```diff
@@ -133,28 +132,19 @@ def populate_row_id_csv(row_ids: Set[str], temp_file):
...
+        :param target_table: The table name with which to format the target filename.
         :return: The gs-path of the uploaded file.
         """
         storage_client = storage.Client(project=self.bucket_project, credentials=self.gcp_creds)
         bucket = storage_client.bucket(self.bucket)
-        blob = bucket.blob(local_filename)
-        blob.upload_from_filename(local_filename)
         target_filename = self._format_filename(table=target_table)
+        blob = bucket.blob(target_filename)
+        blob.upload_from_file(local_file, rewind=True)
```
aherbst-broad, line 147: ❓ What does `rewind=True` mean here?

aherbst-broad (approved):

scripts/hca-utils/hca_utils/utils.py
raaidbroad: ah we put this here when there was a single file open operation, and we wanted to reset the file marker (idk the correct term) back to the beginning of the file; guess it wouldn't be necessary now since we open it in a separate operation to read/upload. Can remove.

PR #44, opened 2021-01-22
Quote data dir var to avoid a syntax error

aherbst-broad:
## Why
The `inject-file-ids` workflow script is dying due to a syntax error in our BigQuery syntax:
```
Syntax error: Expected end of input but got ":" at [5:9]
```
Reproing this in the bq console shows that we're not properly escaping the gcs data dir variable.

## This PR
* Escapes the gcs data dir variable


raaidbroad (approved):
yup that'll do it, good find

quazi-broad (approved):
pretty straightforward

PR #46, opened 2021-01-25
ra dspdc 1477 bad data dir

raaidbroad:
## Why
File ingest is broken, need to fix it.

## This PR
Uses stagingPrefix instead of gcsPrefix to point to correct path of data files so that the file ingest works properly.


aherbst-broad (approved):
LGTM, I think you tested this extensively.

quazi-broad (approved):
looks pretty straightforward

PR #50, opened 2021-02-19
Deduplicate ingest requests by file UUID instead of CRC32C

ThePletch:
We were previously encountering requests for different files being erroneously deduplicated, since crc32c only allows the standard assumptions for a hash (i.e. two identical files will have the same hash, but different files will not necessarily have different hashes). Using the UUID associated with a file instead will ensure that we encounter no erroneous duplicates within the next few millennia.

### DIFF
--- a/transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
+++ b/transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
@@ -225,13 +225,13 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
   ): (String, Msg) = {
     val targetPath = descriptor
       .read[String]("file_name")
-    val contentHash = descriptor
-      .read[String]("crc32c")
+    val contentId = descriptor
+      .read[String]("file_id")
 
-    contentHash -> Obj(
+    contentId -> Obj(
       Str("source_path") -> Str(s"$inputPrefix/data/$targetPath"),
       Str("target_path") -> Str(
-        s"/$contentHash/${targetPath.substring(targetPath.lastIndexOf("/") + 1)}"
+        s"/$contentId/${targetPath.substring(targetPath.lastIndexOf("/") + 1)}"
       )
     )
   }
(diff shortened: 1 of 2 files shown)


transformation/src/test/scala/org/broadinstitute/monster/hca/HcaPipelineBuilderSpec.scala
```diff
@@ -182,19 +182,19 @@ class HcaPipelineBuilderSpec extends AnyFlatSpec with Matchers with PipelineSpec
...
+    val exampleId = "my-file-id"
     val exampleDescriptor = JsonParser.parseEncodedJson(
       json = s"""
                 | {
                 |    "file_name": "a-directory/sub_directory/file-id_file-version_filename.json",
-                |    "file_id": "my-file-id",
+                |    "file_id": "$exampleId",
                 |    "file_version": "my-file-version",
-                |    "crc32c": "$exampleHash",
+                |    "crc32c": "54321zyx",
```
quazi-broad, line 192: will we continue to get the crc32c in the descriptor json?
was the entire purpose of this for deduping?

quazi-broad (approved):
lgtm

transformation/src/test/scala/org/broadinstitute/monster/hca/HcaPipelineBuilderSpec.scala
ThePletch: i don't think i understand the various steps here enough to comment. we'll get the crc32c hash from the HCA regardless, since it's a required field in the file descriptor schema, but i don't know what we're doing with it besides (formerly) using it to dedupe.

raaidbroad: @quazi-broad none of the source data has any defined purpose for us; we made a previous decision to just use one of the hashes for this step.

transformation/src/main/scala/org/broadinstitute/monster/hca/HcaPipelineBuilder.scala
```diff
@@ -225,13 +225,13 @@ object HcaPipelineBuilder extends PipelineBuilder[Args] {
   ): (String, Msg) = {
     val targetPath = descriptor
       .read[String]("file_name")
-    val contentHash = descriptor
-      .read[String]("crc32c")
+    val contentId = descriptor
```
raaidbroad, line 228: Good find, this should be unique across files!

raaidbroad (approved):
Awesome! Clean and precise, good update.

PR #51, opened 2021-02-19
Escape invalid characters in generated dataflow job name

ThePletch:


orchestration/templates/stage-data.yaml
```diff
@@ -83,7 +83,7 @@ spec:
           {{- with .Values.dataflow }}
           - --project={{ .project }}
           - --region={{ .region }}
-          - --jobName=hca-stage-metadata-{{ $stagingPrefix }}
+          - --jobName=hca-stage-metadata-{{ regexReplaceAll "[^-a-z0-9]+" $stagingPrefix "-" }}
```
aherbst-broad, line 86: Where does regexReplaceAll come from?

ThePletch: revisiting this, it comes from helm, which i mistakenly thought was accessible from argo

raaidbroad (approved):
Regex looks sane, again per usual argo things we don't have a good way to know if this will truly work till we try so let 'er rip imo

orchestration/templates/stage-data.yaml
ThePletch: update: helm is accessible from argo. the syntax here was incorrect for function invocations, however. those need to be wrapped in hyphens in addition to double braces. updated.

aherbst-broad (approved):

PR #59, opened 2021-03-01
Add sensor to trigger egress validation when HCA import completes in Argo

ThePletch:
* Adds a sensor that periodically scans through archived (i.e. completed) workflows in the target Argo instance, triggering the  egress validation pipeline for any completed HCA import workflows
* Introduces env-level environment variable config to our Dagster instances via our Helm chart.
* Adds a helper to generate an Argo workflows API client, but doesn't add it as a resource, since sensors can't access resources

### DIFF
--- a/orchestration/dagster_orchestration/hca_orchestration/contrib/argo_workflows.py
+++ b/orchestration/dagster_orchestration/hca_orchestration/contrib/argo_workflows.py
@@ -1,10 +1,14 @@
+from __future__ import annotations  # this lets us annotate functions in class C that return an instance of C
+
+from typing import Any, Callable, Dict, Generator, Optional
+
 from argo.workflows.client import ApiClient as ArgoApiClient,\
                                   ArchivedWorkflowServiceApi,\
                                   Configuration as ArgoConfiguration
-from argo.workflows.client.models import V1alpha1Workflow
+from argo.workflows.client.models import V1alpha1Workflow, V1alpha1WorkflowList
 
 
-def generate_argo_archived_workflows_client(host_url, access_token):
+def generate_argo_archived_workflows_client(host_url: str, access_token: str) -> ArchivedWorkflowServiceApi:
     return ArchivedWorkflowServiceApi(
         api_client=ArgoApiClient(
             configuration=ArgoConfiguration(host=host_url),
@@ -13,22 +17,22 @@ def generate_argo_archived_workflows_client(host_url, access_token):
 
 
 class ArgoArchivedWorkflowsClientMixin:
-    def __init__(self, *args, argo_url, access_token, **kwargs):
+    def __init__(self, *args, argo_url: str, access_token: str, **kwargs):
         super().__init__(*args, **kwargs)
         self.argo_url = argo_url
         self.access_token = access_token
         self._client = None
 
-    def client(self):
+    def client(self) -> ArchivedWorkflowServiceApi:
         if not self._client:
             self._client = generate_argo_archived_workflows_client(self.argo_url, self.access_token)
 
         return self._client
 
-    def list_archived_workflows(self):
-        self._pull_paginated_results(self.client().list_archived_workflows)
+    def list_archived_workflows(self) -> Generator[V1alpha1Workflow, None, None]:
+        return self._pull_paginated_results(self.client().list_archived_workflows)
 
-    def _pull_paginated_results(self, api_function):
+    def _pull_paginated_results(self, api_function: Callable[[Optional[str]], V1alpha1WorkflowList]) -> Generator[V1alpha1Workflow, None, None]:
         results = api_function()
 
         for result in results.items:
@@ -47,19 +51,20 @@ class ExtendedArgoWorkflow(ArgoArchivedWorkflowsClientMixin):
         self._workflow = workflow
         self._inflated = False
 
-    def inflate(self):
+    def inflate(self) -> ExtendedArgoWorkflow:
         if not self._inflated:
             self._workflow = self.client().get_archived_workflow(self.metadata.uid)
             self._inflated = True
 
         return self
 
-    def params_dict(self):
+    def params_dict(self) -> Dict[str, Any]:
         return {
             param.name: param.value
             for param in self.spec.arguments.parameters
         }
 
+    # proxy pattern - any function calls not defined in this class are passed to the wrapped workflow object
     def __getattr__(self, name):
         return getattr(self._workflow, name)
 
--- a/orchestration/dagster_orchestration/hca_orchestration/contrib/argo_workflows.py
+++ b/orchestration/dagster_orchestration/hca_orchestration/contrib/argo_workflows.py
@@ -1,0 +1,67 @@
+from argo.workflows.client import ApiClient as ArgoApiClient,\
+                                  ArchivedWorkflowServiceApi,\
+                                  Configuration as ArgoConfiguration
+from argo.workflows.client.models import V1alpha1Workflow
+
+
+def generate_argo_archived_workflows_client(host_url, access_token):
+    return ArchivedWorkflowServiceApi(
+        api_client=ArgoApiClient(
+            configuration=ArgoConfiguration(host=host_url),
+            header_name="Authorization",
+            header_value=f"Bearer {access_token}"))
+
+
+class ArgoArchivedWorkflowsClientMixin:
+    def __init__(self, *args, argo_url, access_token, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.argo_url = argo_url
+        self.access_token = access_token
+        self._client = None
+
+    def client(self):
+        if not self._client:
+            self._client = generate_argo_archived_workflows_client(self.argo_url, self.access_token)
+
+        return self._client
+
+    def list_archived_workflows(self):
+        self._pull_paginated_results(self.client().list_archived_workflows)
+
+    def _pull_paginated_results(self, api_function):
+        results = api_function()
+
+        for result in results.items:
+            yield result
+
+        while results.metadata['_continue']:
+            results = api_function(list_option_continue=results.metadata['_continue'])
+
+            for result in results.items:
+                yield result
+
+
+class ExtendedArgoWorkflow(ArgoArchivedWorkflowsClientMixin):
+    def __init__(self, workflow: V1alpha1Workflow, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._workflow = workflow
+        self._inflated = False
+
+    def inflate(self):
+        if not self._inflated:
+            self._workflow = self.client().get_archived_workflow(self.metadata.uid)
+            self._inflated = True
+
+        return self
+
+    def params_dict(self):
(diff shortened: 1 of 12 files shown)


orchestration/dagster_orchestration/hca_orchestration/resources/base.py
```diff
@@ -150,14 +159,9 @@ def google_storage_client(init_context):
...
 def jade_data_repo_client(init_context):
-    # get token for jade, assumes application default credentials work for specified environment
-    credentials, _ = google.auth.default()
-    auth_req = Request()
-    credentials.refresh(auth_req)
-
     # create API client
     config = Configuration(host=init_context.resource_config["api_url"])
-    config.access_token = credentials.token
+    config.access_token = default_google_access_token
```
ThePletch, line 164: needs parentheses

orchestration/dagster_orchestration/hca_orchestration/sensors.py
```diff
@@ -0,0 +1,58 @@
...
+def postvalidate_on_import_complete(_):
+    client = generate_argo_workflows_client()
+
+    workflows = [
+        workflow
+        for workflow
+        in client.list_archived_workflows()
+        if workflow.metadata.name.startswith("import-hca-total")
+        and workflow.status.phase == 'Succeeded'
+    ]
```
ThePletch, line 31: this needs test coverage

orchestration/dagster_orchestration/hca_orchestration/sensors.py
```diff
@@ -0,0 +1,58 @@
...
+                        "post_import_validate": {
+                            "config": {
+                                "gcp_env": os.environ.get("HCA_GCP_ENV"),
+                                "google_project_name": os.environ.get("HCA_GOOGLE_PROJECT"),
+                                "dataset_name": workflow_params['data-repo-name'].removeprefix("datarepo_"),
+                            }
+                        }
+                    }
+                }
+            )
```
ThePletch, line 56: same with this

orchestration/dagster_orchestration/hca_orchestration/contrib/argo_workflows.py
```diff
@@ -0,0 +1,67 @@
+from argo.workflows.client import ApiClient as ArgoApiClient,\
+                                  ArchivedWorkflowServiceApi,\
+                                  Configuration as ArgoConfiguration
+from argo.workflows.client.models import V1alpha1Workflow
+
+
+def generate_argo_archived_workflows_client(host_url, access_token):
```
aherbst-broad, line 7: Can this get a type annotation?

orchestration/dagster_orchestration/hca_orchestration/sensors.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+# TODO use execution context to avoid re-scanning old workflows
+@sensor(pipeline_name="validate_egress")
+def postvalidate_on_import_complete(_):
+    sensor = ArgoHcaImportCompletionSensor(argo_url=os.environ.get("HCA_ARGO_URL"), access_token=default_google_access_token())
+
+    workflows = sensor.successful_hca_import_workflows()
+
+    if any(workflows):
```
aherbst-broad, line 46: Would it make sense to add a filter on workflow age here so we avoid spawning unnecessary validation jobs out of the gate (I'm assuming we can get such a thing from the argo api)?

orchestration/dagster_orchestration/hca_orchestration/contrib/argo_workflows.py
```diff
@@ -0,0 +1,67 @@
+from argo.workflows.client import ApiClient as ArgoApiClient,\
+                                  ArchivedWorkflowServiceApi,\
```
aherbst-broad, line 2: Did this not get added? I can't find the class definition for ArchivedWorkflowServiceApi when I search in this PR

aherbst-broad: Ah nm I see that's coming from a lib dependency.

orchestration/dagster_orchestration/hca_orchestration/sensors.py
```diff
@@ -0,0 +1,50 @@
...
+                            "dataset_name": inflated_workflow.params_dict()['data-repo-name'].removeprefix("datarepo_"),
+                        }
+                    }
+                }
+            }
+        )
+
+
+# TODO use execution context to avoid re-scanning old workflows
+@sensor(pipeline_name="validate_egress")
```
aherbst-broad, line 40: How often will this run?

orchestration/dagster_orchestration/hca_orchestration/contrib/argo_workflows.py
```diff
@@ -0,0 +1,67 @@
...
+        self._client = None
+
+    def client(self):
+        if not self._client:
+            self._client = generate_argo_archived_workflows_client(self.argo_url, self.access_token)
+
+        return self._client
+
+    def list_archived_workflows(self):
+        self._pull_paginated_results(self.client().list_archived_workflows)
```
aherbst-broad, line 29: How performant is this endpoint? ie.,. can get we get all the archived workflows in the sensor polling interval or is there some danger of the sensor lapping itself?

aherbst-broad: Also, what happens when a sensor laps itself? Does dagster drop the evaluation on the floor or will the sensor evaluations queue up?

orchestration/dagster_orchestration/hca_orchestration/contrib/argo_workflows.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def _pull_paginated_results(self, api_function):
+        results = api_function()
+
+        for result in results.items:
+            yield result
+
+        while results.metadata['_continue']:
+            results = api_function(list_option_continue=results.metadata['_continue'])
```
aherbst-broad, line 38: What happens with the sensor when argo inevitably 500/502s and this raises an exception?

aherbst-broad (commented):
Looks good so far, had a few questions.

orchestration/dagster_orchestration/hca_orchestration/sensors.py
ThePletch: by default the cadence is "every thirty seconds," but we can throttle that down as needed

orchestration/dagster_orchestration/hca_orchestration/contrib/argo_workflows.py
ThePletch: argo returned a page of results in about a quarter second when i was testing, so i don't expect delays to impact us compared to something like redcap.

as for a sensor lapping itself, it looks like it'll spin up a container to run it regardless of whether the previous run is complete yet, so it's possible to have multiple simultaneous runs for the same sensor. the design appears to be set up to ensure that no executions get dropped. dagster's by-run-name deduplication avoids this creating duplicate runs, but it's a concern for potentially exhausting resources if things start running unusually long for some reason.

orchestration/dagster_orchestration/hca_orchestration/contrib/argo_workflows.py
ThePletch: the run dies and produces no run requests, but it also won't prevent a future run from picking up the run requests that should have been generated by the prior failed run.

orchestration/dagster_orchestration/hca_orchestration/sensors.py
ThePletch: yeah, we should be able to define an epoch to avoid basically backfilling a huge pile of postvalidations for workflows we don't care about

orchestration/dagster_orchestration/hca_orchestration/contrib/argo_workflows.py
ThePletch: probably worth adding error handling here, but unsurprisingly the docs for the autogenerated python client for argo's api are pretty sparse about errors that can be raised by various actions

orchestration/dagster_orchestration/hca_orchestration/sensors.py
```diff
@@ -0,0 +1,56 @@
...
+from typing import List
+
+from dagster import sensor, RunRequest, SkipReason
+
+from hca_orchestration.contrib.argo_workflows import ArgoArchivedWorkflowsClientMixin, ExtendedArgoWorkflow
+from hca_orchestration.resources.base import default_google_access_token
+
+
+# boundary before which we don't care about any workflows in argo
+ARGO_EPOCH = datetime(2021, 3, 1)
```
aherbst-broad, line 12: 👍

orchestration/dagster_orchestration/hca_orchestration/tests/test_sensors.py
```diff
@@ -0,0 +1,139 @@
...
+from hca_orchestration.contrib.argo_workflows import ExtendedArgoWorkflow, generate_argo_archived_workflows_client
+from hca_orchestration.sensors import ArgoHcaImportCompletionSensor
+
+
+T = TypeVar('T')
+
+
+# helper to turn a list into a generator to help mock generator functions
+def generator(iterable: Iterable[T]) -> Generator[T, None, None]:
+    for obj in iterable:
```
aherbst-broad, line 21: An interesting pattern, I haven't seen this before. 👍

aherbst-broad (approved):
lgtm

orchestration/dagster_orchestration/hca_orchestration/tests/test_sensors.py
ThePletch: it's not necessary, strictly speaking, but it's to make sure we correctly provide a generator when mocking functions that return one (e.g. the 'pull paginated results' helper)

orchestration/dagster_orchestration/hca_orchestration/contrib/argo_workflows.py
ThePletch: checked in with the dagster team in their slack, it turns out the behavior is actually that the sensor interval doesn't start counting until the current evaluation of the sensor completes, so if it takes ten minutes to evaluate, it'll wait thirty seconds after that ten minute run completes before starting another.

orchestration/dagster_orchestration/hca_orchestration/resources/base.py
```diff
@@ -13,6 +13,15 @@
 POLLING_INTERVAL = 5  # seconds
 
 
+def default_google_access_token():
+    # get token for google-based auth use, assumes application default credentials work for specified environment
+    credentials, _ = google.auth.default()
+    auth_req = Request()
+    credentials.refresh(auth_req)
```
raaidbroad, line 20: can probably just pass `Request()` straight to `credentials.refresh` instead of the intermediate var since `auth_req` isn't used elsewhere, right?

raaidbroad (approved):

PR #60, opened 2021-03-02
Ra dspdc 1528 jade client hca utils

raaidbroad:
## Why
We now have a Jade api client in Python, so we should use that instead of hacky reverse-engineering of API metadata!

## This PR
Adds a parameter to HcaUtils for a data repo client, removes hacky API code, and passes the client in to HcaUtils from both the command line tool and the solid that calls HcaUtils.


orchestration/dagster_orchestration/hca_utils/main.py
```diff
@@ -13,6 +17,21 @@ def error(self, message):
         sys.exit(2)
 
 
+def get_api_client(host: str):
+    # get token for jade, assumes application default credentials work for specified environment
+    credentials, _ = google.auth.default()
+    auth_req = Request()
+    credentials.refresh(auth_req)
```
aherbst-broad, line 24: Any reason you can't inline the construction of the empty `Request` object here rather than assigning to the throwaway `auth_req` var? Doesn't look like it's used anywhere  else.

aherbst-broad (approved):
Looks straightforward.

orchestration/dagster_orchestration/hca_orchestration/solids.py
```diff
@@ -79,5 +79,6 @@ def post_import_validate(context) -> DagsterProblemCount:
     """
     Checks if the target dataset has any rows with duplicate IDs or null file references.
     """
-    validator = HcaUtils(context.solid_config["gcp_env"], context.solid_config["google_project_name"], context.solid_config["dataset_name"])
+    validator = HcaUtils(context.solid_config["gcp_env"], context.solid_config["google_project_name"],
+                         context.solid_config["dataset_name"], context.resources.data_repo_client)
```
ThePletch, line 83: python arguments supplied across multiple lines should only have one parameter per line for ease of locating specific parameters

orchestration/dagster_orchestration/hca_utils/main.py
```diff
@@ -13,6 +17,20 @@ def error(self, message):
...
+def get_api_client(host: str) -> RepositoryApi:
+    # get token for jade, assumes application default credentials work for specified environment
+    credentials, _ = google.auth.default()
+    credentials.refresh(Request())
+
+    # create API client
+    config = Configuration(host=host)
+    config.access_token = credentials.token
+    client = ApiClient(configuration=config)
+    client.client_side_validation = False
```
ThePletch, line 29: can these property assignments be done in the constructors instead?

orchestration/dagster_orchestration/hca_utils/utils.py
```diff
@@ -159,72 +157,13 @@ def put_csv_in_bucket(self, local_file, target_table: str) -> str:
...
-
     def get_dataset_id(self) -> str:
         """
         Get the dataset ID of the provided dataset name.
         :return: The dataset id.
         """
 
-        response = self._hit_jade("enumerateDatasets", query={"filter": self.dataset})
+        response = self.data_repo_client.enumerate_datasets(filter=self.dataset)
         return response.json()["items"][0]["id"]
```
ThePletch, line 167: not part of this pr,  but i don't love that the error message we'll get if we try to look up a nonexistent dataset is pretty opaque (i.e. an index error instead of a "not found" error)

orchestration/dagster_orchestration/hca_utils/utils.py
```diff
@@ -236,21 +175,18 @@ def submit_soft_delete(self, target_table: str, target_path: str) -> str:
...
-                        "fileType": "csv",
-                        "path": target_path
-                    },
-                    "tableName": target_table
-                }
-            ]
-        }
-
-        response = self._hit_jade("applyDatasetDataDeletion", body=body, params={"id": dataset_id})
+        response = self.data_repo_client.apply_dataset_data_deletion(id=dataset_id,
```
ThePletch, line 178: consider moving this first argument down to its own line, spaced just one indent level out, and lining up the other arguments to that indent level instead.

ThePletch (approved):
some style and brevity suggestions, but nothing blocking release

orchestration/dagster_orchestration/hca_utils/main.py
raaidbroad: Surprisingly it doesn't look like it

PR #62, opened 2021-03-05
Add new pipeline and sensor to main repository, keep repository in separate file

ThePletch:
* Updates our main repository to include the new pipeline (and the new sensor that triggers it)
* Refactors our repository code to be kept in its own file
* Adds a helper script to handle forwarding ports for Dagit access


ops/helmfiles/dagster/forward_ports.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
aherbst-broad, line 1: Thanks for scripting this up

aherbst-broad (approved):

eparkmonster (approved):

PR #66, opened 2021-03-05
Remove redundant step during PR validation

aherbst-broad:


aherbst-broad:
gcloud is needed to as the HCAValidationPipelineSpec reaches out to GS for test files.

PR #70, opened 2021-03-08
Fix env config, sensor implementation for post-validation pipeline

ThePletch:
## This PR
* Fixes our Dagster sensor configuration and implementation (including a handful of syntax fixes)
* Fixes the helm chart's env injection (with comments to avoid us tripping over the same issue in the future)
* Splits out pipelines so each pipeline gets its own file, then pulls them all into the base pipelines package
  * Splits modes to be distinct across each pipeline as well as each env (which lets us exclude resources not used in a pipeline from their mode)

## Still Todo
* Fix permissions for the Dagster service account. The post-validation pipeline is still failing on Jade interactions due to insufficient permissions. Captured in DSPDC-1633.


orchestration/dagster_orchestration/pyproject.toml
```diff
@@ -13,6 +13,8 @@ dagster-k8s = "^0.10.5"
 data-repo-client = "^1.0.194.post5"
 argo-workflows = "^5.0.0"
 dagster-postgres = "^0.10.9"
+dagster-celery-k8s = "^0.10.9"
```
ThePletch, line 16: no

ops/helmfiles/dagster/helmfile.yaml
```diff
@@ -27,20 +27,28 @@ releases:
...
           # "user code" we deploy to dagster; these are dagster repositories of pipelines
           enabled: true
           deployments:
-            # if you add a new deployment here, make sure it's also reflected in the env files (environment/*.yaml)
-            # so that env vars get propagated to it correctly
             - name: "monster-hca-ingest"
               image:
                 repository: "us.gcr.io/broad-dsp-gcr-public/monster-hca-dagster"
-                tag: {{ requiredEnv "GIT_SHORTHASH" }}
+                tag: {{ requiredEnv "GIT_SHORTHASH" | quote }}
```
ThePletch, line 45: learned the hard way that if we don't do this, certain hex combos will be parsed as numbers in very silly ways (2e34677 is 2 to the 34677th power, for instance)

orchestration/dagster_orchestration/hca_orchestration/sensors.py
```diff
@@ -37,13 +39,20 @@ def generate_run_request(self, workflow: ExtendedArgoWorkflow) -> RunRequest:
                             "dataset_name": inflated_workflow.params_dict()['data-repo-name'].removeprefix("datarepo_"),
                         }
                     }
+                },
+                "resources": {
+                    "data_repo_client": {
+                        "config": {
+                            "api_url": "https://jade.datarepo-dev.broadinstitute.org"
```
ThePletch, line 46: swap this to an env var fetch like the other config fields

orchestration/dagster_orchestration/hca_orchestration/pipelines/__init__.py
```diff
@@ -0,0 +1,8 @@
+from hca_orchestration.pipelines.stage_data import stage_data
+from hca_orchestration.pipelines.validate_egress import validate_egress
+
+
+__all__ = [
+    stage_data,
+    validate_egress,
+]
```
ThePletch, line 8: not sure if this needs an explanatory comment - this controls what happens when you call `from hca_orchestration.pipelines import *`

generally this file lets us import the pipelines themselves directly from `hca_orchestration.pipelines` instead of needing to go into their individual implementation files.

orchestration/dagster_orchestration/hca_orchestration/sensors.py
aherbst-broad: Is that a todo, and should it be a small-bite ticket?

orchestration/dagster_orchestration/hca_orchestration/sensors.py
```diff
@@ -2,14 +2,16 @@
...
+
 from dagster import sensor, RunRequest, SkipReason
 
 from hca_orchestration.contrib.argo_workflows import ArgoArchivedWorkflowsClientMixin, ExtendedArgoWorkflow
 from hca_orchestration.resources.base import default_google_access_token
 
 
 # boundary before which we don't care about any workflows in argo
-ARGO_EPOCH = datetime(2021, 3, 1)
+ARGO_EPOCH = datetime(2021, 3, 1, tzinfo=tzlocal())
```
aherbst-broad, line 14: I'd change this to 3/15 at this point, to avoid more reprocessing.

aherbst-broad (approved):

raaidbroad (approved):

orchestration/dagster_orchestration/hca_orchestration/sensors.py
ThePletch: nope, already added to this pr

PR #75, opened 2021-03-16
Fix ordering of params

aherbst-broad:
## Why

The Jade client needs to be the second param to HcaManage.

## This PR
* Changes the param ordering
* TODO: This should get caught in a unit test; there is a large refactoring PR in flight that should land before we engage in the needed refactoring to make HcaManage testable against this case (specifically we should be able to assert that BQ was called with the correct project name and dataset which isn't possible atm without significant mocking)


ThePletch (approved):

ThePletch (approved):
♻️

raaidbroad (approved):

PR #76, opened 2021-03-17
Preconfigure Slack resource with its token

ThePletch:
This will let us run Slack-dependent pipelines from the playground, since we won't have easy access to their token. It also removes the step that tries to configure the resource with a channel, which isn't an option it supports.

### DIFF
--- a/orchestration/dagster_orchestration/hca_orchestration/resources/data_repo.py
+++ b/orchestration/dagster_orchestration/hca_orchestration/resources/data_repo.py
@@ -24,7 +24,7 @@ def base_jade_data_repo_client(init_context: InitResourceContext):
 
 
 @configured(base_jade_data_repo_client)
-def jade_data_repo_client(config):
+def jade_data_repo_client(_config):
     return {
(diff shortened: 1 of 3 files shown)


orchestration/dagster_orchestration/hca_orchestration/resources/data_repo.py
```diff
@@ -24,7 +24,7 @@ def base_jade_data_repo_client(init_context: InitResourceContext):
 
 
 @configured(base_jade_data_repo_client)
-def jade_data_repo_client(config):
+def jade_data_repo_client(_config):
```
aherbst-broad, line 27: ❓ What does the _ prefix mean? Unused?

ThePletch: yup, generally i prepend an underscore if the variable is required in the signature but not used in the body

aherbst-broad (approved):

raaidbroad (approved):

PR #77, opened 2021-03-17
Fix cache ID generation to let us cache Poetry dependencies properly

ThePletch:
It looks like our call to `hashFiles` isn't generating its path properly, which is preventing Github from giving caches proper names (and thus blocking us from properly caching our dependencies). I've updated it to use a relative path, since the documentation indicates that all files passed to "hashFiles" are relative to the Github workspace.

### DIFF
--- a/.github/workflows/validate-python.yaml
+++ b/.github/workflows/validate-python.yaml
@@ -27,7 +27,7 @@ jobs:
         cache-name: cache-poetry
       with:
         path: ~/.cache/pypoetry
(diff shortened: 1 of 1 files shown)


quazi-broad (approved):
lgtm

ThePletch:
> lgtm; it'd be worth it to bump this PR with a noop commit and verify the caching works in the action that gets triggered, prior to merge to avoid another PR loop.

the glitch in this case was from name generation, so i felt confident that it was good to go when i checked the caching step in the original commit's build and saw it correctly generating a hash for the dependencies file

PR #78, opened 2021-03-17
Clean up some bugs/typos in Dagster code

ThePletch:
* Fixes misplaced parentheses in HCA manage
* Distinguishes between HCA google project and data repo google project (@raaidbroad could use your guidance to make sure this approach is the correct one)
* Pulls in secrets config in a few Dagster components that we originally missed


raaidbroad (approved):
yup correct distinction for the google projects

PR #83, opened 2021-03-23
Stringify numbers for environment variable config

ThePletch:
## Why

Helm gets very upset if you try and assign a numeric value to an environment variable instead of a string representing that number.

## This PR


aherbst-broad (approved):
Tested locally, this works.

PR #85, opened 2021-03-25
[DSPDC-1669] Add script to import staging area to the prod dataset

aherbst-broad:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1669)
I put together a small bash script to save some potential fat-fingering of dataset IDs, load tags and other things.

## This PR
* Puts the script into source control for potential reuse

### DIFF
--- a/scripts/release_tools/import_prod.sh
+++ b/scripts/release_tools/import_prod.sh
@@ -1,9 +1,18 @@
 #!/usr/bin/env bash
 
-# Imports a given staging area to the production DCP2 dataset
+# Imports a given staging area to the production DCP2 dataset. This script expects to be run from the
+# scripts/release_tools subdirectory of the HCA repo.
 # Prerequisites:
-#   * a working installation of the argo CLI tool (https://argoproj.github.io/argo-workflows/cli/)
-#   * access to the broad-dsp-monster-hca-prod google project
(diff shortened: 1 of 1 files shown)


quazi-broad (approved):
nice, this should be helpful!

scripts/release_tools/import_prod.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+
+# Imports a given staging area to the production DCP2 dataset
+# Prerequisites:
+#   * a working installation of the argo CLI tool (https://argoproj.github.io/argo-workflows/cli/)
+#   * access to the broad-dsp-monster-hca-prod google project
```
ThePletch, line 6: disambiguation nitpicking:
* access to the project how? with the currently logged in gcloud cli account? do the credentials have to be application default?
* you'll also need `gcloud` and  `kubectl` installed
* `gdate` comes from GNU coreutils, so make sure to mention that people need that installed (`brew install coreutils`)
* this script relies on relative paths, so it will fail if run in a working directory other than the directory it's located in. we should either note that in the prerequisites or alter it to handle its working directory automatically.

ThePletch (approved):
some notes about more thorough documentation, but the actual script looks good

PR #86, opened 2021-03-25
Downgrade to Dagster 0.10.9 until Helm chart supports SA configuration

ThePletch:
## Why

See info in commented-out resources file. We're waiting on a fix for [this Dagster Helm issue](https://github.com/dagster-io/dagster/issues/3934).


aherbst-broad (approved):
:( a shame, but makes sense so we don't block code getting out.

ops/helmfiles/dagster/helmfile.yaml
```diff
@@ -55,7 +55,6 @@ releases:
             - name: monster-dagster-secrets
       - userDeployments:
           # "user code" we deploy to dagster; these are dagster repositories of pipelines
-          enabled: true
```
ThePletch, line 58: (not really part of the downgrade, but this line is redundant with default values)

eparkmonster (approved):

PR #90, opened 2021-03-26
E2E pipeline test with real resources

aherbst-broad:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1562)
We'd like to be able to run the `stage-data` pipeline against real test data with a real beam runner.

## This PR
* Adds a `test_stage_data_local_e2e` test that runs the `stage-data` pipeline against our `ebi_small` dataset using a local beam runner. 
* Adds a `e2e` pytest marker so we can selectively exclude this test. It's pretty quick locally, but there are likely scenarios where we don't want to reach out to GS when testing.


orchestration/dagster_orchestration/hca_orchestration/environments/stage_data_local_e2e.yaml
```diff
@@ -0,0 +1,15 @@
+resources:
+  beam_runner:
+    config:
+      target_class: "hca-transformation-pipeline"
+      working_dir: "../../"
+solids:
+  clear_staging_dir:
+    config:
+      staging_bucket_name: broad-dsp-monster-hca-dev-temp-storage
+      # staging_prefix_name: DETERMINED AT RUNTIME
```
aherbst-broad, line 10: We need to create the staging_prefix at runtime so I'm excluding it here and then putting it in the dict we're passing to the pipeline running machinery at test-time.

@ThePletch you've been working in depth w/dagster configuration patterns so I'm interested in your thoughts on this one.

orchestration/dagster_orchestration/pytest.ini
```diff
@@ -2,3 +2,4 @@
 console_output_style = progress
 env_files =
 	.env.test
+markers = e2e: marks tests as an end-to-end test (deselect with '-m "not e2e"')
```
aherbst-broad, line 5: I've added this marker so we can segment out potentially slow-running E2E tests and not have them run locally if we so choose. In reality this is actually quick (~20 seconds), so it should be ok to keep in our default local test suite.

.github/workflows/validate-python.yaml
```diff
@@ -45,6 +50,6 @@ jobs:
       working-directory: ${{ github.workspace }}/orchestration/dagster_orchestration
       if: always() && steps.dependencies.outcome == 'success' # run all three checks, even if a prior check fails
     - name: Run test suite
-      run: poetry run pytest
+      run: poetry run pytest -v
```
raaidbroad, line 53: I'm having some trouble finding what this flag/arg does; anyone care to enlighten me? `-v` usually means version to me which I'm guessing we're not doing here, and my 5 minutes of googling has not turned up anything that makes sense

orchestration/dagster_orchestration/hca_orchestration/environments/stage_data_local_e2e.yaml
raaidbroad: If this works, it seems like an okay solution; is this something we'd want to generate during runtime for real runs? If so, we might not want it to be a part of the config. If this is just for testing, then I think it's okay

raaidbroad (approved):
Some good refactors, looks good to me overall, left a question/thought or two.

raaidbroad (changes requested):
Oh wait actually; please add type hints/annotations!

.github/workflows/validate-python.yaml
ThePletch: this is the 'verbose' flag, it controls how much gets output to the console

orchestration/dagster_orchestration/hca_orchestration/tests/test_pipelines.py
```diff
@@ -14,25 +18,47 @@ def config_path(relative_path):
...
+        )
+        return self.run_pipeline(pipeline, config_dict, mode, *execution_args, **execution_kwargs)
+
+    @pytest.mark.e2e
+    def test_stage_data_local_e2e(self):
+        test_id = f'test-{uuid.uuid4()}'
+
+        config = load_yaml_from_globs(config_path('stage_data_local_e2e.yaml'))
+        config['solids']['clear_staging_dir']['config']['staging_prefix_name'] = f'local-stage-data/{test_id}'
+        config['solids']['pre_process_metadata']['config']['staging_prefix_name'] = f'local-stage-data/{test_id}'
```
ThePletch, line 42: instead of manually altering the config, we can generate a [preset](https://docs.dagster.io/_apidocs/presets) from the base config and use the `with_additional_config` method to generate a version with this extra config by passing the new config as a dict. example below:
```python
        base_test_preset = PresetDefinition.from_files("test_stage_data_e2e", config_files=config_path('stage_data_local_e2e.yaml'), mode='local')
        test_preset = base_test_preset.with_additional_config({'solids': {'clear_staging_dir': {'config': {'staging_prefix_name': f'local-stage-data/{test_id}'}}, 'pre_process_metadata': {'config': {'staging_prefix_name': f'local-stage-data/{test_id}'}}}})
```

then you can run the pipeline with `self.run_pipeline(stage_data, preset="test_stage_data_e2e")`, and it'll pull in the config you specified, though you'll need to make some alterations to make `config_dict` and `mode` optional fields, since `mode` is mutually exclusive with `preset`.

ThePletch: i haven't played around with this much, so it's possible that this may not be exactly the syntax/structure for this, but i do think using presets here is the right approach.

ThePletch (changes requested):
some suggestions for dealing with dynamically generated config/cascading config files.

orchestration/dagster_orchestration/hca_orchestration/tests/test_pipelines.py
ThePletch: alternatively, if we find that the "preset" construct is more of a hindrance than a help, dagster ships with a tool to merge configs into each other: `from dagster.utils.merger import deep_merge_dicts`

.github/workflows/validate-python.yaml
aherbst-broad: Yeah I added this because I wanted a bit of feedback during the test cycle, happy to remove if folks feel otherwise.

orchestration/dagster_orchestration/hca_orchestration/tests/test_pipelines.py
aherbst-broad: We spoke about this offline but sharing here for anyone else following: the preset approach is a good idea but won't fit this use-case. Specifically, we need to be able to determine the test_id and use for downstream test assertions. The preset approach forces us to set the test_id at pipeline-definition-time which precludes using it downstream in assertions. So, I've opted for the dict deep merge approach.

orchestration/dagster_orchestration/hca_manage/diff_dirs.py
```diff
@@ -12,7 +12,8 @@
 logging.basicConfig(level=logging.INFO)
 
 
-def run(project: str, source_bucket: str, source_prefix: str, target_bucket: str, target_prefix: str) -> None:
+def diff_dirs(project: str, source_bucket: str, source_prefix: str,
+              target_bucket: str, target_prefix: str) -> tuple[dict[str, int], dict[str, int]]:
```
ThePletch, line 16: my preference for function indentation is either "all on one line" or "one line per argument" with no in between, for the sake of being able to easily identify where arguments start and stop.

orchestration/dagster_orchestration/hca_orchestration/environments/stage_data_local_e2e.yaml
```diff
@@ -0,0 +1,13 @@
+resources:
+  beam_runner:
+    config:
+      target_class: "hca-transformation-pipeline"
+      working_dir: "../../"
```
ThePletch, line 5: any way around using a relative path in our config?

orchestration/dagster_orchestration/hca_orchestration/tests/test_pipelines.py
```diff
@@ -1,37 +1,71 @@
...
+def config_path(relative_path: str) -> str:
+    path: str = file_relative_path(
         __file__, os.path.join("../environments/", relative_path)
     )
+    return path
 
 
 class PipelinesTestCase(unittest.TestCase):
-    def run_pipeline(self, pipeline, config_name, *execution_args, **execution_kwargs):
+    def run_pipeline(self, pipeline: PipelineDefinition, config_name: str, extra_config: dict[str, Any] = {},
```
ThePletch, line 25: once you've pulled in my typing changes you should be able to extend `DagsterConfigDict` in `hca_orchestration.support.typing` to produce a more restrictive type for pipeline config dicts

ThePletch (approved):
generally looks good, couple of places where i have questions about formatting or overly permissive type signatures.

raaidbroad (approved):

PR #91, opened 2021-03-26
Add migration script

aherbst-broad:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1679)
We should be able to execute dagster migrations via a script.

## This PR
* Adds a script to execute migrations

### DIFF
--- a/ops/helmfiles/dagster/migrate.sh
+++ b/ops/helmfiles/dagster/migrate.sh
@@ -1,0 +1,33 @@
+#!/usr/bin/env bash
+
+# Migrates our dagster cluster DB schema
+# (taken from https://docs.dagster.io/deployment/guides/kubernetes/how-to-migrate-your-instance)
+
+# Get the names of Dagster's Dagit and Daemon Deployments created by Helm
+export DAGIT_DEPLOYMENT_NAME=`kubectl -n dagster get deploy \
+    --selector=component=dagit -o jsonpath="{.items[0].metadata.name}"`
+export DAEMON_DEPLOYMENT_NAME=`kubectl -n dagster get deploy \
+    --selector=component=dagster-daemon -o jsonpath="{.items[0].metadata.name}"`
+
+# Save each deployment's replica count to scale back up after migrating
+export DAGIT_DEPLOYMENT_REPLICA_COUNT=`kubectl -n dagster get deploy \
+    --selector=component=dagit -o jsonpath="{.items[0].status.replicas}"`
+export DAEMON_DEPLOYMENT_REPLICA_COUNT=`kubectl -n dagster get deploy \
+    --selector=component=dagster-daemon -o jsonpath="{.items[0].status.replicas}"`
+
+echo "scaling down..."
+kubectl -n dagster scale deploy $DAGIT_DEPLOYMENT_NAME --replicas=0
+kubectl -n dagster scale deploy $DAEMON_DEPLOYMENT_NAME --replicas=0
+
+export HELM_DAGSTER_RELEASE_NAME="monster"
+
+echo "applying migration..."
+helm template $HELM_DAGSTER_RELEASE_NAME dagster/dagster \
+    --set "migrate.enabled=true" \
+    --show-only templates/job-instance-migrate.yaml \
+    | kubectl apply -f -
+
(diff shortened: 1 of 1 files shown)


ops/helmfiles/dagster/migrate.sh
```diff
@@ -0,0 +1,33 @@
...
+
+# Save each deployment's replica count to scale back up after migrating
+export DAGIT_DEPLOYMENT_REPLICA_COUNT=`kubectl -n dagster get deploy \
+    --selector=component=dagit -o jsonpath="{.items[0].status.replicas}"`
+export DAEMON_DEPLOYMENT_REPLICA_COUNT=`kubectl -n dagster get deploy \
+    --selector=component=dagster-daemon -o jsonpath="{.items[0].status.replicas}"`
+
+echo "scaling down..."
+kubectl -n dagster scale deploy $DAGIT_DEPLOYMENT_NAME --replicas=0
+kubectl -n dagster scale deploy $DAEMON_DEPLOYMENT_NAME --replicas=0
```
ThePletch, line 20: since we're doing identical things to each of these components, could we pull these into "spin down"/"spin up" functions?

ops/helmfiles/dagster/migrate.sh
```diff
@@ -0,0 +1,33 @@
...
+export DAEMON_DEPLOYMENT_NAME=`kubectl -n dagster get deploy \
+    --selector=component=dagster-daemon -o jsonpath="{.items[0].metadata.name}"`
+
+# Save each deployment's replica count to scale back up after migrating
+export DAGIT_DEPLOYMENT_REPLICA_COUNT=`kubectl -n dagster get deploy \
+    --selector=component=dagit -o jsonpath="{.items[0].status.replicas}"`
+export DAEMON_DEPLOYMENT_REPLICA_COUNT=`kubectl -n dagster get deploy \
+    --selector=component=dagster-daemon -o jsonpath="{.items[0].status.replicas}"`
+
+echo "scaling down..."
```
ThePletch, line 18: design preference: i like to output status messages in shell scripts to stderr, to make it easier to parse script output if we decide to let other scripts consume this one in the future.

ThePletch (approved):
two style suggestions, but this looks good functionally

raaidbroad (approved):
looks like their guide 👍🏾

ops/helmfiles/dagster/migrate.sh
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export HELM_DAGSTER_RELEASE_NAME="monster"
+
+# Get the names of Dagster's Dagit and Daemon Deployments created by Helm
+DAGIT_DEPLOYMENT_NAME=$(kubectl -n dagster get deploy \
+  --selector=component=dagit -o jsonpath="{.items[0].metadata.name}")
+export DAGIT_DEPLOYMENT_NAME
+DAEMON_DEPLOYMENT_NAME=$(kubectl -n dagster get deploy \
+  --selector=component=dagster-daemon -o jsonpath="{.items[0].metadata.name}")
```
ThePletch, line 25: you can combine the `export` statement with the assignment statement

ops/helmfiles/dagster/migrate.sh
```diff
@@ -0,0 +1,48 @@
...
+function msg() {
+  echo $1 >&2
+}
+
+function scale_down() {
+  kubectl -n dagster scale deploy $1 --replicas=0
+}
+
+function scale_up() {
+  kubectl -n dagster scale deploy $1 --replicas=$2
```
ThePletch, line 15: i was thinking you could also compute the deployment name/replica count here, maybe returning the replica count from the function with an `echo`.

ThePletch (commented):
there are some more steps you can take to reduce duplication in the code here. lmk if you want to pair on some of my suggestions.

ops/helmfiles/dagster/migrate.sh
aherbst-broad: Hmm, shellcheck [does not like that](https://github.com/koalaman/shellcheck/wiki/SC2155), I might be missing something with that rule though.

PR #100, opened 2021-04-12
[DSPDC-1698] Staging dataset creation solid

aherbst-broad:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1698).
Continuing to build out the HCA ingest in dagster, the next step is to create a staging dataset in the Monster HCA project that will hold rows for ingest to Jade.

## This PR
* Creates the solid to do this, along with needed configuration schema
*  Creates a bigquery client resource
* Updates pipeline tests to reflect the presence of this solid and assert that it is creating the dataset


orchestration/dagster_orchestration/hca_orchestration/environments/stage_data_local_e2e.yaml
```diff
@@ -2,6 +2,10 @@ resources:
   beam_runner:
     config:
       target_class: "hca-transformation-pipeline"
+  load_tag:
+      config:
+        load_tag_prefix: "fake_load_tag_prefix"
```
quazi-broad, line 7: Is this just a placeholder until we finish building out all of HCA ingest?

orchestration/dagster_orchestration/hca_orchestration/tests/test_resources.py
```diff
@@ -19,13 +19,13 @@
...
-        {
-            'test_resource': config
-        }
+            {
+                'test_resource': resource_def,
+            },
+            DagsterInstance.get(),
+            {
+                'test_resource': config
+            }
```
ThePletch, line 28: very unimportant: why the indent change?

quazi-broad (approved):
changes lgtm, I was trying to pull up the [relevant ticket](https://broadinstitute.atlassian.net/browse/DSDPC-1698) but I get an "access denied"...

orchestration/dagster_orchestration/hca_orchestration/pipelines/stage_data.py
```diff
@@ -37,4 +44,4 @@
     mode_defs=[prod_mode, local_mode, test_mode]
 )
 def stage_data() -> None:
-    pre_process_metadata(start=clear_staging_dir())
+    create_staging_dataset(pre_process_metadata(start=clear_staging_dir()))
```
ThePletch, line 47: probably worth changing the keyword argument to a positional argument here, just for the sake of consistency

orchestration/dagster_orchestration/hca_orchestration/resources/__init__.py
```diff
@@ -13,5 +15,8 @@
     local_beam_runner,
     local_storage_client,
     noop_data_repo_client,
-    test_beam_runner
+    test_beam_runner,
+    noop_bigquery_client,
+    bigquery_client,
+    load_tag
```
ThePletch, line 21: do we think we're reaching a point where shoving these into the top-level class is getting cumbersome? when do we think we'll reach that point, if not?

orchestration/dagster_orchestration/hca_orchestration/resources/bigquery.py
```diff
@@ -0,0 +1,18 @@
+from google.cloud.bigquery import Dataset, Client
+from dagster import resource, InitResourceContext
+from hca_orchestration.contrib.google import get_credentials
+
+
+@resource
+def bigquery_client(init_context: InitResourceContext) -> Client:
+    return Client(credentials=get_credentials())
```
ThePletch, line 8: do we anticipate needing to configure multiple sets of credentials at any point, and would it help to add support for that here?

orchestration/dagster_orchestration/hca_orchestration/resources/load_tag.py
```diff
@@ -0,0 +1,18 @@
+from dagster import resource, Field, StringSource, InitResourceContext, BoolSource
+from datetime import datetime
+
+
+@resource({
+    "load_tag_prefix": Field(StringSource),
+    "append_timestamp": Field(BoolSource),
```
ThePletch, line 7: nit: we can just use `String` and `Boolean` (`Bool`? not sure)  here, the `Source` suffix is only important if we want to load their values from env vars

orchestration/dagster_orchestration/hca_orchestration/tests/test_pipelines.py
```diff
@@ -67,7 +64,7 @@ def test_stage_data_local_e2e(self):
                     'config': {
                         'staging_prefix_name': f'local-stage-data/{test_id}'
                     }
-                }
+                },
```
ThePletch, line 67: :D

orchestration/dagster_orchestration/hca_orchestration/tests/test_resources.py
```diff
@@ -47,7 +47,7 @@ class DataflowBeamRunnerTestCase(unittest.TestCase):
         'DATAFLOW_SUBNET_NAME': 'snubnet',
         'GCLOUD_REGION': 'ec-void1',
         'DATAFLOW_WORKER_MACHINE_TYPE': 'most-expensive-4',
-        'DATAFLOW_STARTING_WORKERS': '2',   # these are marked as ints behind the scenes, but
+        'DATAFLOW_STARTING_WORKERS': '2',  # these are marked as ints behind the scenes, but
```
ThePletch, line 50: wondering if we have some kind of drift in linter settings here  - this doesn't get enforced on my end, and it was indented that way to align with the second line of the comment below it

orchestration/dagster_orchestration/hca_orchestration/tests/test_resources.py
```diff
@@ -57,3 +57,38 @@ class DataflowBeamRunnerTestCase(unittest.TestCase):
...
+            run_config={
+                "resources": {"load_tag": {"config": {
+                    "load_tag_prefix": "fake_prefix",
+                    "append_timestamp": False
+                }}}
+            },
+            mode_def=ModeDefinition(resource_defs={"load_tag": load_tag})
+        )
+
+        self.assertEqual(result.output_value(), "fake_prefix")
```
ThePletch, line 94: testing a resource by affixing it to a solid seems really roundabout. why can't we just test it by initializing it with `initialize_resource` and asserting things about its properties/methods?

orchestration/dagster_orchestration/hca_orchestration/solids/stage_data.py
```diff
@@ -57,3 +59,36 @@ def pre_process_metadata(context: AbstractComputeExecutionContext) -> Nothing:
...
+    required_resource_keys={"bigquery_client", "load_tag"},
+    config_schema={
+        "staging_bq_project": String,
+        "staging_dataset_prefix": String,
+        "staging_table_expiration_ms": Int
+    },
+    input_defs=[InputDefinition("start", Nothing)],
+    output_defs=[OutputDefinition(name="staging_dataset_name", dagster_type=String)]
+)
+def create_staging_dataset(context: AbstractComputeExecutionContext) -> String:
```
ThePletch, line 74: nit: this should return a `str` rather than a `String`  - `String`  here is the dagster config property data type, but we're returning a bare python string

ThePletch (approved):
whole bunch of nitpicks and suggestions for consistency and simplicity. zero complaints about the logic.

orchestration/dagster_orchestration/hca_orchestration/environments/stage_data_local_e2e.yaml
aherbst-broad: This is for the E2E pipeline testing, so I think it should remain a "fake" tag.  I'll give this file a clearer name.

aherbst-broad:
@quazi-broad link updated, I fat fingered the project prefix.

orchestration/dagster_orchestration/hca_orchestration/tests/test_resources.py
aherbst-broad: Autopep8 commit hook fired, not sure why it picked that up.

orchestration/dagster_orchestration/hca_orchestration/resources/__init__.py
aherbst-broad: 🤔 So, I foresee us having the following resources:
* Slack
* Load tag
* BigQuery (staging area)
* BigQuery (jade) (this may be redundant if we use the jade project ID at runtime instead of baking it at init time)
* Jade API client
* Beam runners
* Google storage

Each of those will have different flavors depending on the mode. So, yeah, to your point, I think this is an opportune time to refactor. Perhaps we can split them into a hierarchy of environment packages?

orchestration/dagster_orchestration/hca_orchestration/resources/bigquery.py
aherbst-broad: No. All of these are running as the same SA which we'll give the appropriate permissions.

orchestration/dagster_orchestration/hca_orchestration/tests/test_resources.py
aherbst-broad: I must've touched it manually, putting it back and running the linter doesn't result in a modification by the linter. I'll change it back to avoid unnecessary history.

orchestration/dagster_orchestration/hca_orchestration/tests/test_resources.py
aherbst-broad: Gah, yeah you're right, and there is a perfectly usable pattern right above it. Fixing.

orchestration/dagster_orchestration/hca_orchestration/tests/test_resources.py
aherbst-broad: Turns out this was intellij being helpful, I'm looking for a way to have it respect our linter settings.

PR #106, opened 2021-04-26
Ra dspdc 1688 refactor hca manage

raaidbroad:
## Why
The code for the hca manage CLI has gotten rather difficult to read and the usability is questionable. We ought to break it up into discrete commands.
[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1688)

## This PR
Refactors the CLI into separate commands (does not break up the underlying god object yet).


orchestration/dagster_orchestration/hca_manage/check.py
```diff
@@ -0,0 +1,49 @@
...
+    args = parser.parse_args(arguments)
+    host = data_repo_host[args.env]
+
+    if args.remove:
+        if query_yes_no("Are you sure?"):
+            check_data(args, host, parser, remove=True)
+        else:
+            print("No removal attempted.")
+    else:
+        check_data(args, host, parser)
```
ThePletch, line 30: can we move this confirmation dialog into `check_data` directly, so we ensure that it gets called for all invocations?

orchestration/dagster_orchestration/hca_manage/dataset.py
```diff
@@ -0,0 +1,53 @@
...
+        else:
+            # TODO change to logger?
+            print("No deletion attempted.")
+    elif args.create:
+        if query_yes_no("This will create a dataset. Are you sure?"):
+            create_dataset(args, host)
+
+
+def remove_dataset(args: argparse.Namespace, host: str) -> JobId:
+    hca = HcaManage(environment=args.env, data_repo_client=get_api_client(host=host))
```
ThePletch, line 43: possible to factor out initializing the `HcaManage` client to a helper method?

ThePletch (approved):
This looks good. One thing you might consider for a CLI like this with multiple subcategories is using [subcommands](https://docs.python.org/3/library/argparse.html#sub-commands) to unify them under a central command. That'd let you keep actions like version checking and env selection at the base level to avoid repetition in subcommands, while still letting you fully separate the different parts of the CLI.

ThePletch:
Also, it looks like you'll need to update the references to `hca_manage.main` in various Dagster pipelines.

aherbst-broad (approved):

PR #107, opened 2021-04-26
Configure resources via config files, move common config to resources

ThePletch:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1706)

Configuring our Dagster pipelines using environment variables meant that:
* There was no assurance that all necessary config would be provided for a pipeline
* Configuration lived in our Helm config, making it hard to find and difficult to cross-reference
* All configuration lived in a single, flat dictionary, which made it hard to tease out what config was linked to what functionality

Additionally, we were seeing some tedium from needing to frequently repeat ourselves in cases where multiple solids in a pipeline shared some identical configuration settings.

### New stuff (major features/refactors)
* Sets up a structure for configuring individual resources with simple, flat YAML files separated by resource and environment
* Introduces the `preconfigure_for_mode` function to facilitate configuring a resource with these files. This function will do the following when our Dagster code first initializes (including in CI):
  * Raise an exception if a required setting is not defined
  * Trigger a warning if an unexpected setting is defined
* Reworks existing resources to use `preconfigure_for_mode`
* Introduces the `dev` mode to several pipelines, to allow us to separately load and use prod and dev settings for the same external services.
* Refactors some existing resources to take new settings as part of their config (e.g. the slack resource now expects a destination channel to be passed in config)

### Refactors/cleanups
* Moves the test config YAML files in the `hca_orchestration/environments` folder to `hca_orchestration/tests/environments`
* Finds collections of config settings that were being used throughout pipelines (e.g. `gcp_env` and `google_project_name` and pulls them out into a resource required by the solids to eliminate duplication. We may be able to add helper functions to these resources in the future to further simplify solids.
* Moves existing config-only resources into the `hca_orchestration.resources.config` package, where all other config resources live.


orchestration/dagster_orchestration/hca_orchestration/config/resources/dataflow_beam_runner/global.yaml
```diff
@@ -0,0 +1,9 @@
+image_name: us.gcr.io/broad-dsp-gcr-public/hca-transformation-pipeline
+image_version: latest
+kubernetes_service_account: monster-dagster
+subnet_name: blarg
```
ThePletch, line 4: this turned up as thus far undefined during testing. will try to hunt down the expected value.

orchestration/dagster_orchestration/hca_orchestration/solids/stage_data.py
```diff
@@ -2,28 +2,25 @@
 
 from dagster import solid, InputDefinition, Nothing, String, Int, OutputDefinition
 from dagster.core.execution.context.compute import AbstractComputeExecutionContext
-
 from google.cloud.bigquery import Dataset
 
-
 STAGING_BUCKET_CONFIG_SCHEMA = {
```
ThePletch, line 7: delete

quazi-broad (approved):
Thanks for the walkthrough, it was very helpful to understand all the changes here.
It's nice to see all the red from the things pulled out and put into a common config!

eparkmonster (approved):

orchestration/dagster_orchestration/hca_orchestration/config/preconfiguration_loader.py
```diff
@@ -0,0 +1,91 @@
...
+        keys_present = set(config.keys())
+        permitted_keys = self.required_keys | self.optional_keys
+        missing_keys = self.required_keys - keys_present
+        extra_keys = keys_present - permitted_keys
+
+        if any(missing_keys):
+            missing_keys_str = ", ".join(key for key in missing_keys)
+            raise ValueError(
+                f"Missing expected preconfigured field(s) in configuration files for {self.name}. "
+                f"Config files do not define these required fields:\n{missing_keys_str}\n"
```
aherbst-broad, line 47: 👍 I like this error message, dagster does a good job of providing actionable msgs like this when misconfigurations occur and this is keeping with that pattern.

orchestration/dagster_orchestration/hca_orchestration/config/preconfiguration_loader.py
```diff
@@ -0,0 +1,91 @@
...
+                f"(config dir: {self.config_file_directory})"
+            )
+
+        if any(extra_keys):
+            extra_keys_str = ", ".join(key for key in extra_keys)
+            warnings.warn(
```
aherbst-broad, line 53: Fascinating, I did not know about the `warnings` package until now.

orchestration/dagster_orchestration/hca_orchestration/resources/__init__.py
```diff
@@ -7,16 +7,16 @@
...
-    jade_data_repo_client,
-    live_slack_client,
-    local_beam_runner,
-    local_storage_client,
-    noop_data_repo_client,
-    test_beam_runner,
-    noop_bigquery_client,
-    bigquery_client,
-    load_tag
+    'console_slack_client',
```
aherbst-broad, line 10: ❓ why quote all of these?

orchestration/dagster_orchestration/hca_orchestration/solids/stage_data.py
aherbst-broad: Is this meant to be a todo?

orchestration/dagster_orchestration/hca_orchestration/tests/support/filesystem.py
```diff
@@ -0,0 +1,21 @@
+from contextlib import contextmanager
+import os
+import shutil
+from typing import Iterator
+
+
+# works as a temporary directory, but with a fixed name
+@contextmanager
+def EphemeralNamedDirectory(dirname: str, parent_directory: str) -> Iterator[str]:
```
aherbst-broad, line 9: It's surprising that python does not provide this out of the box with TemporaryDirectory.

aherbst-broad (approved):
Looks good. We should look into upstreaming this, it seems like a useful pattern for dagster in general.

orchestration/dagster_orchestration/hca_orchestration/resources/__init__.py
ThePletch: i mentioned this briefly in the overview, but it's actually against python spec to pass in the values directly here. you're meant to specify the names of each object being added to `__all__` rather than list the objects themselves.

there's some low-level python explanation for this that i don't fully understand, it has to do with being able to reliably retrieve the objects in question, which may not be possible if you pass their values into `__all__`.

also, it makes mypy sad if you don't make them strings.

orchestration/dagster_orchestration/hca_orchestration/tests/support/filesystem.py
ThePletch: I suspect it's something to do with being unable to make a temporary directory securely (which is one of the guarantees the `tempfile` module makes) if it's made with a fixed name.

PR #108, opened 2021-04-28
Add default scopes to GCloud tokens

ThePletch:
## Why

Our access tokens aren't valid to Jade anymore because they have no associated scopes. This change configures tokens with the four scopes (basically just "identify ourselves") that Jade expects for its API calls.

It also includes an unrelated change to the `forward_ports.sh` script to also handle pointing us at the correct Kubernetes cluster.


aherbst-broad (approved):
lgtm

eparkmonster (approved):

PR #109, opened 2021-04-29
Bump Dagster helm chart to newest version

ThePletch:
Does what it says on the tin. Changes include minor usability improvements for the Dagit dashboard.

Generally, we want to try and bump the Helm chart version in lockstep with bumping the Python dependency version.


aherbst-broad:
Should we bump the python versions as well? I see we are on 0.11.2 in the[ pyproject.toml](https://github.com/DataBiosphere/hca-ingest/blob/master/orchestration/dagster_orchestration/pyproject.toml)

ThePletch:
> Should we bump the python versions as well? I see we are on 0.11.2 in the[ pyproject.toml](https://github.com/DataBiosphere/hca-ingest/blob/master/orchestration/dagster_orchestration/pyproject.toml)

The version in the toml is a minimum version - we're currently on 0.11.6 in poetry.lock. I could definitely be convinced that we should pin the version directly in the toml rather than depend on the lock file, though.

aherbst-broad:
Yeah, I see no reason _not_ to pin; I think we want to keep the helm + toml in lockstep rather than let them potentially drift.

ThePletch:
It is still a manual process to bump the dependency version, since we've checked in poetry.lock. This is mostly a matter of what we think is the most clear.

aherbst-broad:
@ThePletch , right, why not change `dagster = "^0.11.2"'` to `dagster = "0.11.6"` in the pyproject.toml as part of this pr?

raaidbroad (approved):
pending the version pin drew suggested, lgtm

aherbst-broad (approved):
lgtm

PR #112, opened 2021-04-30
Ra dspdc 1688 refactor god object

raaidbroad:
## Why
God object bad. Readable files good.
[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1688)

## This PR
Breaks the large manage.py file/object into multiple files/objects with more specific goals, updates calls that reference the object, and breaks up tests too.

### DIFF
--- a/orchestration/dagster_orchestration/hca_manage/bq_managers.py
+++ b/orchestration/dagster_orchestration/hca_manage/bq_managers.py
@@ -30,7 +30,7 @@ class BQRowManager(_BQRowDataclass, ABC):
         pass
 
     @abstractmethod
-    def check_or_remove_rows(self, soft_delete: bool = False) -> int:
+    def check_or_delete_rows(self, soft_delete: bool = False) -> int:
         pass
 
     def _hit_bigquery(self, query: str) -> set[str]:
@@ -90,7 +90,7 @@ class DanglingFileRefManager(BQRowManager):
         """
         return self._hit_bigquery(query)
 
-    def check_or_remove_rows(self, soft_delete: bool = False) -> int:
+    def check_or_delete_rows(self, soft_delete: bool = False) -> int:
         """
         Check for any entities with project_id values that do not have a corresponding entry in the projects
         table
@@ -160,12 +160,14 @@ class DuplicatesManager(BQRowManager):
 
         return self._hit_bigquery(query)
 
-    def check_or_remove_rows(self, soft_delete: bool = False) -> int:
+    def check_or_delete_rows(self, soft_delete: bool = False) -> int:
         """
         Check and print the number of duplicates for each table in the dataset.
         :return: Number of duplicate rows to soft delete
         """
-        return self._check_or_delete_rows(self.get_all_table_names, self.get_rows, soft_delete=soft_delete,
+        return self._check_or_delete_rows(self.get_all_table_names,
+                                          self.get_rows,
+                                          soft_delete=soft_delete,
                                           issue="duplicate rows")
 
     def get_all_table_names(self) -> set[str]:
@@ -195,7 +197,7 @@ class NullFileRefManager(BQRowManager):
 
         return self._hit_bigquery(query)
 
-    def check_or_remove_rows(self, soft_delete: bool = False) -> int:
+    def check_or_delete_rows(self, soft_delete: bool = False) -> int:
         """
         Check/remove and print the number of null file references for each table in the dataset that has a `file_id`
         column.
--- a/orchestration/dagster_orchestration/hca_manage/check.py
+++ b/orchestration/dagster_orchestration/hca_manage/check.py
@@ -3,7 +3,6 @@ from dataclasses import dataclass
 import logging
 from typing import Optional
 
-from cached_property import cached_property
 from data_repo_client import RepositoryApi
 
 from hca_manage import __version__ as hca_manage_version
@@ -65,7 +64,7 @@ class CheckManager:
     dataset: str
     data_repo_client: RepositoryApi
 
-    @cached_property
+    @property
     def soft_delete_manager(self) -> SoftDeleteManager:
         return SoftDeleteManager(environment=self.environment,
                                  dataset=self.dataset,
@@ -96,9 +95,9 @@ class CheckManager:
         :return: A named tuple with the counts of rows to soft delete
         """
         logging.info("Processing...")
-        duplicate_count = self.duplicate_manager.check_or_remove_rows()
-        null_file_ref_count = self.null_file_ref_manager.check_or_remove_rows()
-        dangling_proj_refs_count = self.dangling_file_ref_manager.check_or_remove_rows()
+        duplicate_count = self.duplicate_manager.check_or_delete_rows()
+        null_file_ref_count = self.null_file_ref_manager.check_or_delete_rows()
+        dangling_proj_refs_count = self.dangling_file_ref_manager.check_or_delete_rows()
         logging.info("Finished.")
         return ProblemCount(
             duplicates=duplicate_count,
@@ -113,8 +112,8 @@ class CheckManager:
         :return: A named tuple with the counts of rows to soft delete
         """
         logging.info("Processing, deleting as we find anything...")
-        duplicate_count = self.duplicate_manager.check_or_remove_rows(soft_delete=True)
-        null_file_ref_count = self.null_file_ref_manager.check_or_remove_rows(soft_delete=True)
+        duplicate_count = self.duplicate_manager.check_or_delete_rows(soft_delete=True)
+        null_file_ref_count = self.null_file_ref_manager.check_or_delete_rows(soft_delete=True)
         logging.info("Skipping any rows with dangling project refs, manual intervention required")
         logging.info("Finished.")
         return ProblemCount(
--- a/orchestration/dagster_orchestration/hca_manage/soft_delete.py
+++ b/orchestration/dagster_orchestration/hca_manage/soft_delete.py
@@ -34,8 +34,10 @@ def run(arguments: Optional[list[str]] = None) -> None:
 
 
 def soft_delete(args: argparse.Namespace, host: str) -> None:
-    hca = SoftDeleteManager(environment=args.env, data_repo_client=get_api_client(host=host),
-                            project=args.project, dataset=args.dataset)
+    hca = SoftDeleteManager(environment=args.env,
+                            data_repo_client=get_api_client(host=host),
+                            project=args.project,
+                            dataset=args.dataset)
     hca.soft_delete_rows(args.path, args.target_table)
 
 
@@ -113,4 +115,4 @@ class SoftDeleteManager:
             )
         )
 
-        return response.id  # type: ignore # data repo client has no type hints, since it's auto-generated
+        return JobId(response.id)  # type: ignore # data repo client has no type hints, since it's auto-generated


orchestration/dagster_orchestration/hca_manage/soft_delete.py
```diff
@@ -24,6 +35,92 @@ def run(arguments: Optional[list[str]] = None) -> None:
...
+    bucket_project: str = field(init=False)
+    bucket: str = field(init=False)
+
+    def __post_init__(self) -> None:
+        self.filename_template = f"sd-{self.project}-{self.dataset}-{uuid.uuid4()}-{{table}}.csv"
+        self.bucket_project = {"prod": "mystical-slate-284720",
+                               "dev": "broad-dsp-monster-hca-dev"}[self.environment]
+        self.bucket = f"broad-dsp-monster-hca-{self.environment}-staging-storage"
+
+    @cached_property
```
aherbst-broad, line 59: I don't think this needs to be cached, it's used exactly once.

orchestration/dagster_orchestration/hca_manage/check.py
```diff
@@ -38,12 +48,251 @@ def check_data(args: argparse.Namespace, host: str, parser: argparse.ArgumentPar
...
     if remove:
         hca.remove_all()
     else:
         hca.check_for_all()
+
+
+@dataclass
+class CheckManager(SoftDeleteManager):
+
+    @cached_property
```
aherbst-broad, line 65: I don't think this needs to be cached, it's used exactly once.

aherbst-broad (commented):
Looking good, had a few questions and I think tests are failing as well.

orchestration/dagster_orchestration/hca_manage/check.py
```diff
@@ -1,9 +1,19 @@
 import argparse
-from typing import Optional
+import csv
+from dataclasses import dataclass
+import logging
+import os
+from typing import Callable, Optional, TextIO
+
+from cached_property import cached_property
```
ThePletch: i screwed up in pioneering this pattern, but in the event that we want to cache properties, `cached_property` is now part of python core, under `functools`

orchestration/dagster_orchestration/hca_manage/check.py
```diff
@@ -38,12 +48,251 @@ def check_data(args: argparse.Namespace, host: str, parser: argparse.ArgumentPar
...
+                       data_repo_client=get_api_client(host=host))
 
     if remove:
         hca.remove_all()
     else:
         hca.check_for_all()
+
+
+@dataclass
+class CheckManager(SoftDeleteManager):
```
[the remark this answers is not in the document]
ThePletch: i agree with this thought and want to expand on it - this class feels like it was decomposed by topic (i.e. "stuff to check and validate rows") instead of by purpose ("fetching rows," "validating rows," "marking rows," etc). it's handling fetching records, running validations on them, and marking them as soft-deleted, as well as utility functions like collecting values into a CSV and running queries in BigQuery. i think we need to work out what individual things this class does and separate out functionality in addition to purpose.

orchestration/dagster_orchestration/hca_manage/check.py
```diff
@@ -38,12 +48,251 @@ def check_data(args: argparse.Namespace, host: str, parser: argparse.ArgumentPar
...
+        :return: A set of table names.
+        """
+        query = f"""
+        SELECT table_name
+        FROM `{self.project}.datarepo_{self.dataset}.INFORMATION_SCHEMA.TABLES` WHERE table_type = "VIEW"
+        """
+
+        return self._hit_bigquery(query)
+
+    @staticmethod
```
ThePletch, line 244: when i'm decomposing objects, static methods are always a signal to me that more things need to be decomposed. since they don't relate to a class's properties by definition, they're typically unrelated to the purpose of a particular class. i think `populate_row_id_csv` is a good example of that.

orchestration/dagster_orchestration/hca_manage/check.py
```diff
@@ -38,12 +48,251 @@ def check_data(args: argparse.Namespace, host: str, parser: argparse.ArgumentPar
...
+        Create a csv locally with one column filled with row ids to soft delete.
+        :param row_ids: A set of row ids to soft delete.
+        :param temp_file: a temporary file to pass in
+        :return: The filename of the created csv.
+        """
+        sd_writer = csv.writer(temp_file, delimiter=",", quotechar='"', quoting=csv.QUOTE_MINIMAL)
+
+        sd_writer.writerows([[rid] for rid in row_ids])
+
+    def _process_rows(
```
ThePletch, line 256: while we're in here, i'd like to suggest that we reword methods like this - the word "process" to me feels a lot like the word "do" or "change" - it's very ambiguous without heavy context.

orchestration/dagster_orchestration/hca_manage/common.py
```diff
@@ -1,11 +1,24 @@
 import argparse
+from dataclasses import dataclass
 import sys
 from typing import NoReturn
 
 from data_repo_client import ApiClient, Configuration, RepositoryApi
 
 from hca_orchestration.contrib.google import default_google_access_token
 
+JobId = str
```
ThePletch, line 10: this should have been changed to a subclass upstream - did you rebase recently?

orchestration/dagster_orchestration/hca_manage/check.py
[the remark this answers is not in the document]
raaidbroad: Hmm there are a bunch of attributes that they share so I figured it'd be simpler to just inherit and extend. I'm open to changing it but this was simple and maintained most of the code in a copy-pastable format haha

orchestration/dagster_orchestration/hca_manage/dataset.py
```diff
@@ -40,14 +47,71 @@ def run(arguments: Optional[list[str]] = None) -> None:
...
+    hca = DatasetManager(environment=args.env, data_repo_client=get_api_client(host=host))
     return hca.create_dataset(
         dataset_name=args.dataset_name,
         billing_profile_id=args.billing_profile_id,
         schema_path=args.schema_path
     )
+
+
+@dataclass
+class DatasetManager:
```
ThePletch, line 64: if i can suggest an alternative pattern here, consider approaching this less as a set of service objects and more as a set of data objects bundled with operations you can perform on those objects. for this one, it'd be less `DatasetManager('prod').delete_dataset('name', ...)` and instead `Dataset('prod', 'name', ...).delete()`

orchestration/dagster_orchestration/hca_manage/snapshot.py
```diff
@@ -50,5 +58,84 @@ def create_snapshot(args: argparse.Namespace, host: str) -> JobId:
...
-    hca = HcaManage(environment=args.env, data_repo_client=get_api_client(host=host))
+    hca = SnapshotManager(environment=args.env, data_repo_client=get_api_client(host=host))
     return hca.delete_snapshot(snapshot_name=args.snapshot_name, snapshot_id=args.snapshot_id)
+
+
+@dataclass
+class SnapshotManager:
+    environment: str
+    data_repo_client: RepositoryApi
+    dataset: Optional[str] = None
```
ThePletch, line 69: this seems odd to me. i would expect `dataset` to be mandatory for a child object of datasets like a snapshot.

orchestration/dagster_orchestration/hca_manage/check.py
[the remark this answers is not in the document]
aherbst-broad: I'd like to keep these classes focused and ensure a good separation of concerns, if possible. It's hard to eyeball this, which attributes are shared?

orchestration/dagster_orchestration/hca_manage/soft_delete.py
```diff
@@ -24,6 +35,92 @@ def run(arguments: Optional[list[str]] = None) -> None:
...
+    data_repo_client: RepositoryApi
+    project: Optional[str] = None
+    dataset: Optional[str] = None
+
+    bucket_project: str = field(init=False)
+    bucket: str = field(init=False)
+
+    def __post_init__(self) -> None:
+        self.filename_template = f"sd-{self.project}-{self.dataset}-{uuid.uuid4()}-{{table}}.csv"
+        self.bucket_project = {"prod": "mystical-slate-284720",
```
ThePletch, line 54: let's pull out hardcoded config maps like this into properties of the dataclass that will be passed in. i don't think we should have hardcoded config in this repo at all at this point, but if it's out of scope to remove this in this pr, let's keep it in a top-level constant.

orchestration/dagster_orchestration/hca_manage/soft_delete.py
```diff
@@ -24,6 +35,92 @@ def run(arguments: Optional[list[str]] = None) -> None:
...
+        return hca_google.get_credentials()
+
+    def soft_delete_rows(self, path: str, target_table: str) -> JobId:
+        with open(path, mode="rb") as rf:
+            remote_file_path = self.put_csv_in_bucket(local_file=rf, target_table=target_table)
+            job_id = self._submit_soft_delete(target_table=target_table, target_path=remote_file_path)
+            logging.info(f"Soft delete job for table {target_table} running, job id of: {job_id}")
+            return job_id
+
+    def put_csv_in_bucket(self, local_file: BinaryIO, target_table: str) -> str:
```
ThePletch, line 69: i feel like both the name and docstring for this method should give more information about what the csv is and does in this context

orchestration/dagster_orchestration/hca_manage/soft_delete.py
```diff
@@ -24,6 +35,92 @@ def run(arguments: Optional[list[str]] = None) -> None:
...
+        target_filename = self._format_filename(table=target_table)
+        blob = bucket.blob(target_filename)
+        blob.upload_from_file(local_file)
+
+        filepath = f"gs://{self.bucket}/{target_filename}"
+        logging.info(f"Put a soft-delete file here: {filepath}")
+
+        return filepath
+
+    def get_dataset_id(self) -> str:
```
ThePletch, line 87: this feels like a utility method without a clear link to this service object's stated purpose of soft deletions

ThePletch (changes requested):
it's a massive undertaking to pull apart something as large as `HcaManage` and this alone is a big step toward making it more usable, but i think there's more we can do here. let's try and focus on breaking down the classes by purpose and contained data rather than by topic.

orchestration/dagster_orchestration/hca_manage/check.py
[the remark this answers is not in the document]
ThePletch: usually if i find two objects share a lot of attributes but one's purpose/functionality isn't a superset of the other's, it's a sign i should pull out the common properties into their own class and have both inherit from that new class.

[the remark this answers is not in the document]
raaidbroad: If we decompose this by purpose rather than topic, do we still have a topic-based object that is simply composed of the purpose-based ones? I think it is useful to organize by "this is how we think of each separate tool", so therefore I think it is useful to have a representation of "this is the thing that does the checks we want". I think you both have more evolved thoughts than I on this though and I am very happy to learn why we should do this; I simply want to make sure I understand it 🙂

orchestration/dagster_orchestration/hca_manage/check.py
raaidbroad: Agreed, process is vague and not helpfully descriptive

orchestration/dagster_orchestration/hca_manage/check.py
raaidbroad: So we'd pull this out into a standalone function and put it in common.py or something?

orchestration/dagster_orchestration/hca_manage/common.py
raaidbroad: I saw that it was a class, but since it is just a string, I figured this was simpler/cleaner

orchestration/dagster_orchestration/hca_manage/snapshot.py
raaidbroad: Ah good catch

orchestration/dagster_orchestration/hca_manage/common.py
ThePletch: I covered this briefly during my review, but there's a difference in behavior here - if you just set it as an alias of `str`, then they're interchangeable, so passing a string that isn't a `JobId` to a function that accepts a `JobId` won't raise any alarms to mypy.

If you make it a subclass (we should have that somewhere), it'll ensure that functions that accept/return `JobId` can't be passed a standard string.

orchestration/dagster_orchestration/hca_manage/check.py
ThePletch: potentially, yeah. if you find a lot of static methods that serve a common purpose, you can pull them into their own class.

orchestration/dagster_orchestration/hca_manage/common.py
raaidbroad: WOW today I learned. Will change it back

orchestration/dagster_orchestration/hca_manage/check.py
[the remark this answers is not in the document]
ThePletch: topic-based classes are definitely valuable for command-line tools, and in my mind, should be very lightweight mappings between CLI arguments/flags and pre-existing functionality in more purpose-focused classes. if you've worked in an MVC framework before, think of the topic-based class as the controller and the purpose/data-based class as the model.

orchestration/dagster_orchestration/hca_manage/snapshot.py
raaidbroad: So `dataset` and `data_repo_profile_id` are not required for deletions, that is why they're `Optional`

orchestration/dagster_orchestration/hca_manage/dataset.py
raaidbroad: I'm not sure how I feel about implementing this change; the `create` method has 3 parameters that the `delete` method doesn't have, and the `delete` method considers one of the `create` method parameters to be optional. To my eyes, it makes it less clear 😕

ThePletch: yeah, giving this a second run-through in my head, it'd be possible to rework things to be more data-focused (with the object constructor just taking required parameters and the various methods taking the additional parameters you mentioned), but it'd require a radical enough reworking that i don't think it'll be beneficial at this point.

if we kept going down this route, certain types of function ('create a new X') would be a more natural fit to live elsewhere (e.g. 'create a snapshot' being a function under the dataset object rather than the snapshot object, 'create a dataset' being one of the base-level client operations or something), which i think would resolve some of the parameter disjunction that you're running into here, but the effort involved in planning out that refactor (and accounting for things like the dynamic dataset ID lookup in the current "delete dataset" operation) would be larger than the benefit we'd get from it, i think, and it's something we can do piecemeal later on if this code starts becoming a problem again.

orchestration/dagster_orchestration/hca_manage/snapshot.py
ThePletch: makes sense to me. my worry here going forward is that optional properties instantly double the number of possible combinations of values that a class can have, so it's harder to reason about how a function will behave under all circumstances as more and more optional properties show up, but just the fact that we've split these classes out has drastically reduced that problem already, so it's not nearly as big of a deal.

orchestration/dagster_orchestration/hca_manage/check.py
aherbst-broad: So, does this actually need to be cached? I'm curious why this decorator is being applied here.

aherbst-broad (approved):
I think this is a sensible refactoring

orchestration/dagster_orchestration/hca_manage/bq_managers.py
```diff
@@ -0,0 +1,226 @@
...
+        EXCEPT DISTINCT SELECT rid
+                        FROM (
+                            SELECT MAX(rid) AS rid, fid
+                            FROM ridsOfLatestFids
+                            GROUP BY fid)
+        """
+
+        return self._hit_bigquery(query)
+
+    def check_or_remove_rows(self, soft_delete: bool = False) -> int:
```
ThePletch, line 163: i'm confused by the distinction between `remove` and `delete` here - can we rename this?

orchestration/dagster_orchestration/hca_manage/check.py
ThePletch: confused why this has been marked as resolved - drew's right, i'm unclear on why we want to cache this (and if we do, we should use the python core version of `cached_property`)

orchestration/dagster_orchestration/hca_manage/soft_delete.py
```diff
@@ -24,6 +34,83 @@ def run(arguments: Optional[list[str]] = None) -> None:
 
 
 def soft_delete(args: argparse.Namespace, host: str) -> None:
-    hca = HcaManage(environment=args.env, data_repo_client=get_api_client(host=host),
-                    project=args.project, dataset=args.dataset)
+    hca = SoftDeleteManager(environment=args.env, data_repo_client=get_api_client(host=host),
+                            project=args.project, dataset=args.dataset)
```
ThePletch, line 38: nit: arguments should either all be on one line or each have their own line
```suggestion
    hca = SoftDeleteManager(
        environment=args.env,
        data_repo_client=get_api_client(host=host),
        project=args.project,
        dataset=args.dataset
   )
```

orchestration/dagster_orchestration/hca_manage/soft_delete.py
```diff
@@ -24,6 +34,83 @@ def run(arguments: Optional[list[str]] = None) -> None:
...
+                            "fileType": "csv",
+                            "path": target_path
+                        },
+                        "tableName": target_table
+                    }
+                ]
+            )
+        )
+
+        return response.id  # type: ignore # data repo client has no type hints, since it's auto-generated
```
ThePletch, line 116:
```suggestion
        return JobId(response.id)  # type: ignore # data repo client has no type hints, since it's auto-generated
```

PR #113, opened 2021-05-03
Fix Dagit config resource ref in snapshot slack hooks

ThePletch:
Also moves the secret env config for the Slack resource into our config files and removes an unused import.


aherbst-broad (approved):

quazi-broad (approved):
changes look straightforward

PR #114, opened 2021-05-05
Rename "staging" to "scratch"

aherbst-broad:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1699)
"Staging" is an overloaded term in our infrastructure. It can refer to either:
* An area where external collaborators deposit files
* A monster internal area our pipelines use as a scratchpad
I propose we use the term "scratch" for our internal staging area going forward.

## This PR
* Renames usage of "staging" to "scratch" in our dagster pipeline
* Extracts scratch BQ config to a resource config, it will be needed in the data file load branch so I don't want to have to repeat that solid config


orchestration/dagster_orchestration/hca_orchestration/resources/config/scratch.py
```diff
@@ -0,0 +1,24 @@
...
+    scratch_prefix_name: str
+    scratch_bq_project: str
+    scratch_dataset_prefix: str
+    scratch_table_expiration_ms: int
+
+
+@resource({
+    "scratch_bucket_name": String,
+    "scratch_prefix_name": String,
+    "scratch_bq_project": String,
```
aherbst-broad, line 18: These fields will be needed by the data file load code, so rather than repeating solid config, I'm centralizing in this resource config.

raaidbroad (approved):

ThePletch (approved):

PR #115, opened 2021-05-05
[DSPDC-1699] Datafile Pipeline Buildout

aherbst-broad:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1699)
We want to be able to load data files from HCA staging areas into a target dataset.
This is a checkpoint on the buildout of the datafile pipeline. 

TODO: descriptor file ingest

## This PR
* Implements a diff against the existing HCA dataset using BQ, using that as the control file for the Jade bulk file ingest
* Implements the bulk file ingest to Jade  
   * I've added a workaround for a 500 we encounter when attempting to pull the job results too soon after the job succeeds (this is ticketed with the Jade team)
* Adds tests for the new solids
* Marks the e2e pipeline as skip, for now. We should be able to reactivate it, but there is an issue with re-ingesting the data files from the EBI "small" test dataset (Jade says they already exist). Once that gets worked out we can re-enable


orchestration/dagster_orchestration/hca_orchestration/contrib/data_repo.py
```diff
@@ -0,0 +1,45 @@
...
+    :param api_fn: API method to call
+    :param max_wait_time: Maximum wait time in seconds to poll
+    :param poll_interval_seconds: Interval between polling calls
+    :param retriable_status_codes:  Set of HTTP status codes that indicate we should retry
+    :param args: Args to pass to the API fn
+    :return:
+    """
+    time_waited = 0
+
+    while time_waited < max_wait_time:
```
aherbst-broad, line 33: I would like to coalesce this and this [similar op](https://github.com/DataBiosphere/hca-ingest/blob/master/orchestration/dagster_orchestration/hca_orchestration/solids/data_repo.py#L22), but I am unsure atm how to integrate the exception handling needed here cleanly.

orchestration/dagster_orchestration/hca_orchestration/resources/bigquery.py
```diff
@@ -8,11 +10,6 @@ def bigquery_client(init_context: InitResourceContext) -> Client:
...
 
-class NoopBigQueryClient:
-    def create_dataset(self, dataset: Dataset) -> None:
-        pass
-
-
 @resource
 def noop_bigquery_client(init_context: InitResourceContext) -> Client:
-    return NoopBigQueryClient()
+    return Mock(spec=Client)
```
aherbst-broad, line 15: I am leaning toward ripping out all of our "noop" resources and replacing them with `ResourceDefinition.hardcoded_resource(Mock(spec=<class of the thing being mocked>))`, as needed. I'd be interested in any thoughts folks have on this.

orchestration/dagster_orchestration/hca_orchestration/tests/environments/test_load_hca_local_e2e.yaml
```diff
@@ -1,18 +1,24 @@
...
-      scratch_dataset_prefix: local_e2e
+      scratch_bucket_name: broad-dsp-monster-hca-dev-temp-storage
+      scratch_dataset_prefix: test_load_hca_local_e2e
       scratch_table_expiration_ms: 3600000
+  target_hca_dataset:
+    config:
+      dataset_name: hca_dev_20210405_arh_testing
+      dataset_id: f413bcb3-9ec4-41b0-a75b-1e36b19dc3dc
+      project_id: broad-jade-dev-data
+      billing_profile_id: 390e7a85-d47f-4531-b612-165fc977d3bd
```
aherbst-broad, line 20: Once we figure out the issue I mentioned in the PR description, I'm hopeful we can use this config to run e2e locally.

orchestration/dagster_orchestration/hca_orchestration/solids/load_hca/load_data_files.py
```diff
@@ -1,12 +1,236 @@
...
+    external_config: ExternalConfig = ExternalConfig(
+        ExternalSourceFormat.NEWLINE_DELIMITED_JSON
+    )
+    external_config.source_uris = [
+        f"{_gs_path_from_bucket_prefix(scratch_config.scratch_bucket_name, scratch_config.scratch_prefix_name)}/data-transfer-requests/*"
+    ]
+    external_config.schema = [
+        SchemaField("source_path", "STRING"),
+        SchemaField("target_path", "STRING")
+    ]
```
ThePletch, line 91: I'm wondering if bits like this that configure the BigQuery job could be pulled out into a resource (perhaps a `BigQueryConfigurator`) or at least a separate class. This seems too procedural (and too hardcoded) to be part of the main body of the solid.

orchestration/dagster_orchestration/hca_orchestration/solids/load_hca/load_data_files.py
```diff
@@ -1,12 +1,236 @@
...
+    job_config.destination_format = bigquery.DestinationFormat.NEWLINE_DELIMITED_JSON
+    extract_job: bigquery.ExtractJob = bigquery_client.extract_table(
+        f"{scratch_dataset_name}.{file_load_table_name}",
+        destination_uris=[
+            f"{out_path}"
+        ],
+        job_config=job_config,
+        location="US",
+        project=scratch_config.scratch_bq_project
+    )
```
ThePletch, line 132: i think a decent hunk of this (specifically the hardcoded values and constants) could be moved into the "job builder" resource i suggested earlier

orchestration/dagster_orchestration/hca_orchestration/solids/load_hca/load_data_files.py
```diff
@@ -1,12 +1,236 @@
...
+        'poll_interval_seconds': Int,
+    }
+)
+def _base_check_bulk_file_ingest_job_result(context: AbstractComputeExecutionContext, job_id: JobId) -> Nothing:
+    # we need to poll on the endpoint as a workaround for a race condition in TDR (DR-1791)
+    job_results = api_request_with_retry(
+        context.resources.data_repo_client.retrieve_job_result,
+        context.solid_config['max_wait_time_seconds'],
+        context.solid_config['poll_interval_seconds'],
+        {500},
```
ThePletch, line 183: i'm wondering if we can save ourselves some future pain here by marking other common transient 5xx errors (i.e. 502-504) for retry as well

orchestration/dagster_orchestration/hca_orchestration/solids/load_hca/load_data_files.py
```diff
@@ -1,12 +1,236 @@
...
+        job_id
+    )
+    failed_files = job_results['failedFiles']
+    if failed_files > 0:
+        raise Failure(f"Bulk file load (job_id = {job_id} had failedFiles = {failed_files}")
+
+
+@configured(_base_check_bulk_file_ingest_job_result)
+def check_bulk_file_ingest_job_result(config: DagsterConfigDict) -> DagsterConfigDict:
```
ThePletch, line 192: let's generally avoid directly calling `@configured` in favor of using `preconfigure_for_mode` for passing in all default configs like this

orchestration/dagster_orchestration/hca_orchestration/tests/test_pipelines.py
```diff
@@ -43,7 +43,7 @@ def run_pipeline(
             mode=pipeline_mode
         )
 
-    @pytest.mark.e2e
+    @pytest.mark.skip
```
ThePletch, line 46: you should be able to layer on multiple markers rather than swapping the e2e marker out, right?

ThePletch (commented):
some style/syntactic sugar suggestions, plus some ideas for code organization. my strongest opinion here is that we should factor out some of the bigquery job config setup in the solids into either a separate class or a resource.

orchestration/dagster_orchestration/hca_orchestration/support/typing.py
```diff
@@ -3,7 +3,7 @@
 base can live here, for easy reference and descriptive naming.
 """
 
-from dagster import InputDefinition
+from dagster import InputDefinition, usable_as_dagster_type
```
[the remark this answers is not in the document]
ThePletch: i'm a tad worried that flake8 didn't catch this, on further review

orchestration/dagster_orchestration/hca_orchestration/solids/load_hca/load_data_files.py
aherbst-broad: @ThePletch how would that work for a solid in this context? I'm unsure how I'd pull in the mode name

raaidbroad (approved):

orchestration/dagster_orchestration/hca_orchestration/solids/load_hca/load_data_files.py
ThePletch: hm, good point. it seems like the correct approach here is to use a preset, but we'd need to have a think about best practices for presets if we wanted to do that, as well as consider whether there'd be other downsides.

orchestration/dagster_orchestration/hca_orchestration/config/resources/target_hca_dataset/test.yaml
```diff
@@ -0,0 +1,4 @@
+dataset_name: fake_hca_dataset_name
```
ThePletch, line 1: i don't like the pattern of preconfiguring config resources for tests like this, since the config you're specifying is specific to a particular pipeline run. these are essentially test fixtures, and they probably shouldn't be living alongside application config.

orchestration/dagster_orchestration/hca_orchestration/contrib/bigquery.py
```diff
@@ -0,0 +1,65 @@
...
+        table_name: str,
+        destination: str,
+        bigquery_client: bigquery.client.Client,
+        bigquery_project: str
+) -> bigquery.QueryJob:
+    """
+    Performs a bigquery query using an external table definition, using the supplied
+    schema and source paths. Results are deposited at the supplied GCS destination
+    """
+    job_config = bigquery.QueryJobConfig()
```
ThePletch, line 18: idle q: any reason to namespace these rather than reference them directly?

orchestration/dagster_orchestration/hca_orchestration/contrib/retry.py
```diff
@@ -0,0 +1,40 @@
...
+
+def is_truthy(result: Any) -> bool:
+    """
+    Helper that returns True if the provided result is defined, False if not
+    """
+    return bool(result)
+
+
+def retry(target_fn: Callable[..., Any], max_wait_time_seconds: int,
+          poll_interval_seconds: int, success_predicate: Callable[[Any], bool], *args: str) -> Any:
```
ThePletch, line 20: nit: i prefer keeping function definitions to either a single line or one line per argument, as follows:
```suggestion
def retry(
    target_fn: Callable[..., Any],
    max_wait_time_seconds: int,
    poll_interval_seconds: int,
    success_predicate: Callable[[Any], bool],
    *args: str,
) -> Any:
```

orchestration/dagster_orchestration/hca_orchestration/solids/load_hca/load_data_files.py
```diff
@@ -1,12 +1,230 @@
...
+        'mode': 'NULLABLE',
+        'name': 'target_path',
+        'type': 'STRING',
+    }
+]
+
+
+@solid(
+    required_resource_keys={"bigquery_client", "scratch_config", "storage_client", "target_hca_dataset"},
+    input_defs=[InputDefinition("scratch_dataset_name", HcaScratchDatasetName)],
```
ThePletch, line 38: wanted to confirm again whether the `input_def` here is required - i believe dagster should be able to infer it from the argument and annotation

orchestration/dagster_orchestration/hca_orchestration/solids/load_hca/load_data_files.py
```diff
@@ -1,12 +1,230 @@
...
+                    scratch_dataset_name: HcaScratchDatasetName) -> Iterator[str]:
+    """
+    Determine files to load by joining against the target HCA dataset, then drop the result in our scratch bucket as a
+     "control file" that Jade will use for bulk file ingest
+    :param context: Dagster solid context
+    :param scratch_dataset_name: Name of the "scratch" dataset we are using for this pipeline run
+    :return: Yields a list of control file blob names
+    """
+    target_hca_dataset = context.resources.target_hca_dataset
+    file_load_table_name = 'file_load_requests'
```
ThePletch, line 51: i try to avoid hardcoding values in function definitions, in favor of keeping them in either configs or in top-level constants

orchestration/dagster_orchestration/hca_orchestration/solids/load_hca/load_data_files.py
```diff
@@ -1,12 +1,230 @@
...
+    bigquery_client = context.resources.bigquery_client
+    scratch = context.resources.scratch_config
+
+    _determine_files_to_load(
```
ThePletch, line 55: if we're just tossing away the return value of this job, where is the "determination" being tracked? in the scratch table?

i think we may want a comment to this effect.

ThePletch (commented):
some type annotation questions and one point about how we organize test fixtures

orchestration/dagster_orchestration/hca_orchestration/contrib/retry.py
aherbst-broad: Those should be `Any` I think, will make the change and thread through **kwargs as well.

orchestration/dagster_orchestration/hca_orchestration/contrib/bigquery.py
aherbst-broad: Juggling between the bigquery.client and storage.client types I believe.

orchestration/dagster_orchestration/hca_orchestration/contrib/retry.py
```diff
@@ -0,0 +1,49 @@
+import time
+from typing import Any, Callable, TypeVar
+
+
+class RetryException(Exception):
```
ThePletch, line 5: nit: inherit from `RuntimeError` for user-defined exceptions, inheriting from `Exception` implies a system-level failure

orchestration/dagster_orchestration/hca_orchestration/contrib/retry.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+    Represents a failure after some number of retries
+    """
+    pass
+
+
+def is_truthy(result: Any) -> bool:
```
ThePletch, line 12: type nit: when writing functions that are expected to work for all possible values, use `object` instead of `Any` as the argument type to indicate "all" rather than "whichever"

ThePletch (approved):
couple more type suggestions but nothing to block release.

PR #118, opened 2021-05-10
Make primitive subtypes passable via config

ThePletch:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1752)

## This PR

* Makes the simple primitive subtypes passable in run config, so you can stub them out in Dagit. This lets us run pipelines without needing to necessarily start at the beginning for inputs like these.

### DIFF
--- a/orchestration/dagster_orchestration/hca_manage/common.py
+++ b/orchestration/dagster_orchestration/hca_manage/common.py
@@ -4,6 +4,9 @@ from dataclasses import dataclass
 import sys
 from typing import NoReturn, TextIO
 
+from dagster import make_python_type_usable_as_dagster_type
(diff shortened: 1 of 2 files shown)


ThePletch:
I went VERY far down a rabbit hole of making a decorator to make dataclasses usable like this as well, but i could not get mypy to cooperate, so in the end, i abandoned the project. I don't anticipate us getting much use out of a function like that, regardless.

raaidbroad (approved):

ThePletch:
wait we're not doing this wrong, dagster has no idea how to use `NewType` instances

aherbst-broad (approved):
Ah, I'd been looking for something like this, thanks.

PR #120, opened 2021-05-14
Set dataset region on creation

aherbst-broad:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1762)
We need to provide a region when creating a dataset to ensure it gets created in the `US` multi-region, not `us-central1`.
## This PR
* Threads in the necessary args


ThePletch (approved):
makes sense to me

eparkmonster (approved):

quazi-broad (approved):

PR #125, opened 2021-05-20
Pre-flight validation check wrangler CLI script

eparkmonster:
## Why
Want to be able to check that GS paths exist, the staging areas are sane, and validate JSON files in those directories.
[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1697)

## This PR
Validates that the given directory exists and the following staging areas: (/data, /descriptors, /metadata, /links dirs)
Currently a CLI script, taking in the path and the bucket as arguments.
The script will output a list of invalid JSON and their errors or will output that the file path doesn't exist.


orchestration/dagster_orchestration/hca_manage/validation.py
```diff
@@ -0,0 +1,65 @@
...
+    storage_client = storage.Client()
+    bucket = storage_client.bucket(args.bucket)
+
+    validate_directory(path=args.path + '/data', bucket=bucket)
+    validate_directory(path=args.path + '/descriptors', bucket=bucket)
+    validate_directory(path=args.path + '/links', bucket=bucket)
+    validate_directory(path=args.path + '/metadata', bucket=bucket)
+
+
+run()
```
aherbst-broad, line 65: 🔧 Wrap this in an `if __name__ == __main__` so this script doesn't get run if someone happens to import it from other code ([related SO question](https://stackoverflow.com/questions/419163/what-does-if-name-main-do))

orchestration/dagster_orchestration/hca_manage/validation.py
```diff
@@ -0,0 +1,65 @@
...
+    parser = DefaultHelpParser(description="CLI to manage validate GS path and json files.")
+    parser.add_argument("-p", "--path", help="GS path to validate", required=True)
+    parser.add_argument("-b", "--bucket", help="Bucket the GS path exists in", required=True)
+
+    args = parser.parse_args(arguments)
+
+    storage_client = storage.Client()
+    bucket = storage_client.bucket(args.bucket)
+
+    validate_directory(path=args.path + '/data', bucket=bucket)
```
aherbst-broad, line 59: ⛏️ You could extract the "well-known" folder names to a set and then just for comp over it, something like

```suggestion
    well_known_dirs = { '/data', '/descriptors', '/links', '/metadata' }
    {validate_directory(dir) for dir in well_known_dirs}
```

orchestration/dagster_orchestration/hca_manage/validation.py
```diff
@@ -0,0 +1,65 @@
...
+        logging.error(f"{path} File Path doesn't exist")
+    elif len(invalid_files_in_dir) > 0:
+        logging.error(f"List of invalid json files {invalid_files_in_dir}")
+    else:
+        logging.info('File path and Json are valid')
+
+
+def run(arguments: Optional[list[str]] = None) -> None:
+    parser = DefaultHelpParser(description="CLI to manage validate GS path and json files.")
+    parser.add_argument("-p", "--path", help="GS path to validate", required=True)
```
aherbst-broad, line 51: 🔧 Since we're always given full `gs://` paths (i.e., `gs://foo-bucket/bar/baz`), it'd be helpful to consume those as such here. Since these paths are just URLs, I think we can leverage python's builtin URL parsing functionality and create a helper method to accomplish this.

Something like:

```
from urllib.parse import urlparse

...

@dataclass 
class GsBucketWithPrefix:
    bucket: str
    prefix: str

...

def parse_gs_path(raw_gs_path: str) -> GsBucketWithPrefix:
     url_result = urlparse(raw)
     return GsBucketWithPrefix(url_result.bucket, url_result.prefix)

```

... and then just use that for downstream GS calls.

orchestration/dagster_orchestration/hca_manage/validation.py
```diff
@@ -0,0 +1,65 @@
...
+def validate_directory(path: str, bucket: storage.Client.bucket) -> None:
+    """
+    Validate that the directory in the Google storage bucket is valid
+    :param path: Google stage path name
+    :param bucket: Google storage bucket name
+    """
+    valid_files_in_dir = []
+    invalid_files_in_dir = {}
+    for blob in bucket.list_blobs(prefix=path):
+        # bucket_file = bucket.blob(blob.name)
```
aherbst-broad, line 35: ⛏️ Remove this commented out code.

orchestration/dagster_orchestration/hca_manage/validation.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def validate_json(blob: storage.Blob) -> Optional[Exception]:
+    """
+    Validate that the JSON file blob follows the schema in the describedBy
+    :param blob: JSON file blob to check
+    :return: An exception if JSON is invalid, return None if valid
+    """
+    try:
+        data = json.loads(blob.download_as_string(client=None))
```
aherbst-broad, line 17: ⛏️ Seems like `client=None` is redundant, can you just call it without the optional kwarg? ie.. 

```suggestion
        data = json.loads(blob.download_as_string())
```

orchestration/dagster_orchestration/hca_manage/validation.py
```diff
@@ -0,0 +1,65 @@
...
+        schema_path = data["describedBy"]
+        raw_schema = requests.get(schema_path).json()
```
aherbst-broad, line 19: 🔧 This is going to be expensive if we have to pull down the schema for every json file we validate (there's not that many). 

What if we create a top level method dict that can cache schema bodies and we can refer to that?


```suggestion

class SchemaFetcher:
    def __init__(self):
        self._schema_cache = {}

    def fetch_schema(self, path: str):
        if path in self._schema_cache:
            return self._schema_cache[path]

        raw_schema = requests.get(schema_path).json()
        self._schema_cache[path] = raw_schema
        return raw_schema



...


def validate_json(blob: storage.Blob, schema_fetcher: SchemaFetcher) -> Optional[Exception]:

...

schema = schema_fetcher.fetch_schema(schema_path)
          

```

aherbst-broad (commented):
Looking good, had a few comments.

orchestration/dagster_orchestration/hca_manage/validation.py
```diff
@@ -0,0 +1,90 @@
...
+    bucket: str
+    prefix: str
+
+
+def parse_gs_path(raw_gs_path: str) -> GsBucketWithPrefix:
+    url_result = urlparse(raw_gs_path)
+    return GsBucketWithPrefix(url_result.netloc, url_result.path[1:])
+
+
+def validate_json(blob: storage.Blob, schema_fetcher: Type[SchemaFetcher]) -> Optional[Exception]:
```
aherbst-broad, line 36: ⛏️ This is just a `SchemaFetcher`, not a `Type[SchemaFetcher]`
```suggestion
def validate_json(blob: storage.Blob, schema_fetcher:SchemaFetcher) -> Optional[Exception]:
```

orchestration/dagster_orchestration/hca_manage/validation.py
```diff
@@ -0,0 +1,90 @@
...
+def validate_directory(path: str, bucket: storage.Client.bucket) -> None:
+    """
+    Validate that the directory in the Google storage bucket is valid
+    :param path: Google stage path name
+    :param bucket: Google storage bucket name
+    """
+    valid_files_in_dir = []
+    invalid_files_in_dir = {}
+    for blob in bucket.list_blobs(prefix=path):
+        json_error = validate_json(blob, SchemaFetcher)
```
aherbst-broad, line 62: 🔧 This needs to be a `SchemaFetcher` instance, not the class. I suggest adding a `schema_fetcher: SchemaFetcher` parameter to this function and passing it in.

aherbst-broad, line 62: This should also only validate JSON files; as it stands, this will validate the any data files present which can be multi-gigabyte and are definitely not JSON.

orchestration/dagster_orchestration/hca_manage/validation.py
```diff
@@ -0,0 +1,94 @@
...
+def validate_directory(path: str, bucket: storage.Client.bucket, schema_fetcher: SchemaFetcher) -> None:
+    """
+    Validate that the directory in the Google storage bucket is valid
+    :param path: Google stage path name
+    :param bucket: Google storage bucket name
+    """
+    valid_files_in_dir = []
+    invalid_files_in_dir = {}
+    for blob in bucket.list_blobs(prefix=path):
+        if blob.name[-4:] == 'json':
```
aherbst-broad, line 62: ⛏️ `blob.endswith` would work better here.

aherbst-broad (approved):

orchestration/dagster_orchestration/mypy.ini
```diff
@@ -57,3 +57,6 @@ ignore_missing_imports = True
 
 [mypy-kubernetes.*]
 ignore_missing_imports = True
+
+[mypy-jsonschema.*]
+ignore_missing_imports = True
```
aherbst-broad, line 62: ⛏️ Don't forget to add a newline here.

orchestration/dagster_orchestration/hca_manage/validation.py
```diff
@@ -0,0 +1,94 @@
+from google.cloud import storage
+import logging
+from jsonschema import validate
+import json
+import requests
+from typing import Optional, Dict, Any
+from hca_manage.common import DefaultHelpParser
+from urllib.parse import urlparse
+from dataclasses import dataclass
```
raaidbroad, line 9: nitpick: for python, it is good practice to order imports alphabetically, and in sections. First section is python built-in libraries, like logging, json, etc. second section is third party libraries, like google.cloud, and third section is package-specific imports, like hca_manage. You want to alphabetically order within each section.

raaidbroad (commented):
Do we want any tests for anything here, or would it be tautological at best?

aherbst-broad:
@raaidbroad Tautological, let's hold off.

orchestration/dagster_orchestration/hca_manage/validation.py
aherbst-broad: 🤔 wondering if we can enforce this with autopep8

eparkmonster: Makes sense, I'll also make sure to keep this in mind for future python

raaidbroad (approved):

PR #126, opened 2021-05-20
[DSPDC-1732] Rework release verifier tool

aherbst-broad:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1732)
Not all source paths in a staging area will be loaded; this happens if multiple staging areas contain the same file.
We should check if the desired _target path_ was loaded.

## This PR
* Reworks the verifier script to reflect the above logic
* Adds the option to parallelize the verification of multiple staging areas via multiprocessing


eparkmonster (approved):

quazi-broad (approved):
lgtm, the new logic is a little easier to follow and the comments were helpful.

orchestration/dagster_orchestration/hca_manage/verify_release_manifest.py
```diff
@@ -1,72 +1,96 @@
...
 from dagster_utils.contrib.google import get_credentials
 
 logging.basicConfig(level=logging.INFO, format='%(message)s')
 
 
-def get_expected_load_totals(storage_client: Client, staging_areas: list[str]) -> dict[str, int]:
+@dataclass(frozen=True)
+class PathWithCrc:
+    path: str
+    crc: str
```
raaidbroad, line 34: do we perhaps want to really explicitly call out that this is the crc32c, given the confusion we've experienced before?

orchestration/dagster_orchestration/hca_manage/verify_release_manifest.py
```diff
@@ -1,72 +1,96 @@
...
+
+
+FILE_TYPES = {
```
raaidbroad, line 37: not a big deal at all, but I think we have this list of file types in more than one place (I guess I'm thinking of that enum you added in another PR); do we want some central "source of truth" for something like this or is that overkill?

raaidbroad (commented):
looks good, I think we should be explicit about what term we use to specify the crc32c but am open to pushback

PR #127, opened 2021-05-21
[DSPDC-1776] Fix validation sensor

aherbst-broad:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1776)
Our post-import validation sensor does not work in dev, as it is hardcoded to the `prod` mode. 

## This PR
* Adds an `env` environment variable to our runtime environment variable set.
* Threads this env var through to a new builder method that will then instantiate the sensor with the correct mode
* Adds some light docs
* Updates to the latest dagster


quazi-broad (approved):
changes make sense to me

eparkmonster (approved):

raaidbroad (approved):

PR #129, opened 2021-05-24
Ra dspdc 1768 non file metadata

raaidbroad:
## Why
We've added the file metadata ingest, and we now need to add the non-file metadata ingest to complete the 'load' part of the hca pipeline.
[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1768)

## This PR
Adds `non_file_metadata` under `hca_orchestration/solids/load_hca`, which contains a composite solid called `non_file_metadata_fanout` that maps over all the non-file metadata types with the existing `load_table` solid. The logic is very similar (read: copy-paste-edited) to what Drew added in the `data_files` package. The composite solid is also added to the `load_hca` pipeline.

Deleted specific file or non-file metadata types and created a more generic `MetadataType` and corresponding fanout result type.

`load_table` has been refactored to say "metadata" instead of "file metadata", and logic has been added to point to the right storage location for file vs. non file metadata.

This PR also sneaks in the `.gitattributes` file and marks the poetry.lock file as linguist generated, so that it won't show up in PR diffs.


orchestration/dagster_orchestration/hca_orchestration/solids/load_hca/non_file_metadata/load_non_file_metadata.py
```diff
@@ -0,0 +1,58 @@
...
+    SPECIMEN_FROM_ORGANISM = NonFileMetadataType("specimen_from_organism")
+    LINKS = NonFileMetadataType("links")
+
+
+@solid(
+    output_defs=[
+        DynamicOutputDefinition(name="table_fanout_result", dagster_type=NonFileMetadataTypeFanoutResult)
+    ]
+)
+def ingest_metadata_type(scratch_dataset_name: HcaScratchDatasetName) -> Iterator[NonFileMetadataTypeFanoutResult]:
```
raaidbroad, line 43: this solid is very similar to the one in `load_data_metadata_files`. I think it is small and simple enough that abstracting it out into its own function might take more time/effort than it is worth, plus this keeps the typing very very tight and scoped. If we would prefer to make this a generalized solid, I am happy to do so!

orchestration/dagster_orchestration/hca_orchestration/solids/load_hca/non_file_metadata/typing.py
```diff
@@ -0,0 +1,12 @@
+from typing import NamedTuple
+
+from hca_orchestration.support.typing import HcaScratchDatasetName
+
+
+class NonFileMetadataType(str):
```
raaidbroad, line 6: I made these in parallel to what Drew made for the file metadata types, again to keep stuff tightly typed; if we think it makes sense for it all to fall under the same type, I can change it to that as well.

orchestration/dagster_orchestration/hca_orchestration/solids/load_hca/non_file_metadata/load_non_file_metadata.py
```diff
@@ -0,0 +1,58 @@
...
+    SPECIMEN_FROM_ORGANISM = NonFileMetadataType("specimen_from_organism")
+    LINKS = NonFileMetadataType("links")
+
+
+@solid(
+    output_defs=[
+        DynamicOutputDefinition(name="table_fanout_result", dagster_type=NonFileMetadataTypeFanoutResult)
+    ]
+)
+def ingest_non_file_metadata_type(scratch_dataset_name: HcaScratchDatasetName) -> Iterator[NonFileMetadataTypeFanoutResult]:
```
raaidbroad, line 43: this solid is very similar to what Drew wrote in the file metadata solids; if we think it is worth abstracting out and having just one solid, I can do that. It didn't feel worth the time/effort and this keeps it more tightly type scoped but I am not strongly opinionated on this.

orchestration/dagster_orchestration/hca_orchestration/solids/load_hca/load_table.py
```diff
@@ -9,13 +9,13 @@
...
-from hca_orchestration.solids.load_hca.data_files.typing import FileMetadataType, FileMetadataTypeFanoutResult
 from hca_orchestration.solids.load_hca.poll_ingest_job import check_data_ingest_job_result
-from hca_orchestration.support.typing import HcaScratchDatasetName
+from hca_orchestration.support.typing import HcaScratchDatasetName, MetadataType, MetadataTypeFanoutResult
 
 
 def _diff_hca_table(
-        file_metadata_type: FileMetadataType,
+        metadata_type: MetadataType,
+        is_file_metadata: bool,
```
raaidbroad, line 18: truly only adding this here because we need to be able to point to a different source_path based on whether it is file metadata or non-file metadata

orchestration/dagster_orchestration/hca_orchestration/solids/load_hca/load_table.py
```diff
@@ -29,26 +29,28 @@ def _diff_hca_table(
 
     query = f"""
     SELECT J.datarepo_row_id, S.*, {datarepo_key}
-    FROM {file_metadata_type} S FULL JOIN {target_hca_dataset.project_id}.{fq_dataset_id}.{file_metadata_type} J
+    FROM {metadata_type} S FULL JOIN {target_hca_dataset.project_id}.{fq_dataset_id}.{metadata_type} J
     USING ({primary_key})
     """
     destination = f"{scratch_dataset_name}.{joined_table_name}"
+
+    metadata_path = "file-metadata-with-ids" if is_file_metadata else "metadata"
```
raaidbroad, line 37: here is the switch for file vs nonfile metadata external source paths

orchestration/dagster_orchestration/hca_orchestration/support/typing.py
```diff
@@ -12,3 +14,13 @@ class HcaScratchDatasetName(str):
 
 
 make_python_type_usable_as_dagster_type(HcaScratchDatasetName, DagsterString)
+
+
+class MetadataType(str):
```
raaidbroad, line 19: This more generic type and fanout result type works for file and nonfile metadata, replaces the existing ones (that are now deleted)

orchestration/dagster_orchestration/hca_orchestration/support/typing.py
```diff
@@ -12,3 +14,13 @@ class HcaScratchDatasetName(str):
...
+    pass
+
+
+class MetadataTypeFanoutResult(NamedTuple):
+    scratch_dataset_name: HcaScratchDatasetName
+    metadata_type: MetadataType
+    is_file_metadata: bool = False
```
aherbst-broad, line 26: 🔧 this might be where we express the path in an earlier comment; rather than leaving it up to a function way downstream to care about whether it's file metadata or non-file metadata, we could determine that at the top and simplify the downstream logic.

aherbst-broad (commented):
looking good, had a question about the `is_file_metadata` bool.

orchestration/dagster_orchestration/hca_orchestration/solids/load_hca/load_table.py
raaidbroad: Hmm I'm not sure about the rename; the dataflow pipeline writes the `metadata` output, which includes the file metadata _without_ file ids as well as the rest of the metadata entities, and it seems like pointless processing to move the non-file metadata to another folder and name it differently

aherbst-broad (approved):

eparkmonster (approved):
LGTM

PR #130, opened 2021-05-24
Add dataset policy members on creation

aherbst-broad:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1717)
We want to add relevant dataset policy members at the time of creation to avoid manual setup in swagger.

## This PR
* Adds a `create_dataset_with_policy_members` method to the `hca_manage` class, which creates the dataset, polls on it's creation then adds the specified policy members, along with a test
* Reflects the extracted job polling code from [this dagster-utils ](https://github.com/broadinstitute/dagster-utils/pull/15)PR
* Adds an `_exec_tdr_op` helper method to the dataset manager, which wraps all TDR calls in the dataset script with a handler that looks for 401 responses. If a 401 is received, it outputs a more helpful message rather than the usual traceback spew. 

## TODO
- [x] Update `dagster-utils` version once related PR is merged

### DIFF
--- a/orchestration/dagster_orchestration/hca_manage/common.py
+++ b/orchestration/dagster_orchestration/hca_manage/common.py
@@ -8,13 +8,8 @@ from dagster import make_python_type_usable_as_dagster_type
 from dagster.core.types.dagster_type import String as DagsterString
 
 from dagster_utils.contrib.google import default_google_access_token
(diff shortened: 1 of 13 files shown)


quazi-broad (approved):
changes look good to me

eparkmonster (approved):

PR #132, opened 2021-05-27
Snapshotting improvements

aherbst-broad:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1789)
During the last snapshot process, there were a few pain points:
* No job ID was exposed via slack, making it difficult to debug when the job failed
* The monster team did not have reader access to the generated snapshot, which is good to have for potential debugging

## This PR
* Updates the snapshotting job to expose the job ID
* Improves the slack integration to pass through a "block" layout to make the notifications more readable
* Adds `monster@firecloud.org` to the snapshot readers in prod

### DIFF
--- a/orchestration/dagster_orchestration/hca_manage/snapshot.py
+++ b/orchestration/dagster_orchestration/hca_manage/snapshot.py
@@ -74,7 +74,7 @@ class SnapshotManager:
     def __post_init__(self) -> None:
         self.reader_list = {
             "dev": ["hca-snapshot-readers@dev.test.firecloud.org"],
-            "prod": ["hca-snapshot-readers@firecloud.org"]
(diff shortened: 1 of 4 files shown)


eparkmonster (approved):

quazi-broad (approved):
lgtm.
Just had a question about why the slack notification functions were being put into the pipelines code? My initial reaction would have been to try to create a slack notification resource - is it better practice to keep logging/tracking/notification code in the pipelines for a dagster project?

raaidbroad (approved):

aherbst-broad:
@quazi-broad an interesting idea. I think in this case we wouldn't realize much gain, as this is purely adding some markup that's being passed to the existing slack notification resource, and it's not preventing us from testing at all. I think I prefer to keep it simple for now and we can refactor if it becomes cumbersome.

PR #135, opened 2021-06-02
Ability to query for snapshots

aherbst-broad:
## Why

We often need to query for snapshots, this adds the ability to our hca_manage tool so we don't have to go to the swagger page.

## This PR
* Adds the needed query method which hits the `enumerateSnapshots` endpoint
* Adds back the ability to create a dataset with a pre-baked JSON schema file, which I need for DAP TDR testing.
* Updates arg parsing in the `snapshots.py` file to use subparsers rather than mutually exclusive arg groups, bringing it in line with the approach I took in `dataset.py`.
* Minor cleanups in `dataset.py`

### DIFF
--- a/orchestration/dagster_orchestration/hca_manage/common.py
+++ b/orchestration/dagster_orchestration/hca_manage/common.py
@@ -1,6 +1,7 @@
 import argparse
 import csv
 from dataclasses import dataclass
+import logging
 import functools
 import sys
 from typing import Any, Callable, NoReturn, TextIO, TypeVar, cast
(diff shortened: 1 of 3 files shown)


orchestration/dagster_orchestration/hca_manage/common.py
```diff
@@ -25,16 +37,8 @@ def has_problems(self) -> bool:
...
-    "prod": "https://jade-terra.datarepo-prod.broadinstitute.org/",
-    "real_prod": "https://data.terra.bio/"
-}
-
-data_repo_profile_ids = {
-    "dev": "390e7a85-d47f-4531-b612-165fc977d3bd",
-    "prod": "db61c343-6dfe-4d14-84e9-60ddf97ea73f"
-}
+def setup_cli_logging_format():
+    logging.basicConfig(level=logging.INFO, format='%(message)s')
```
aherbst-broad, line 41: I am centralizing our logging format for CLI tools here, and ditching the timestamps + loglevel as I find them a bit noisy.

eparkmonster (approved):
LGTM

raaidbroad (approved):

PR #136, opened 2021-06-03
[DSPDC-1793] Add billing profile CLI

aherbst-broad:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1793)

We create billing profiles from time to time, we should be able to do this from the CLI outside of the Jade swagger.

## This PR
* Adds a simple CLI job to create billing profiles

### DIFF
--- a/orchestration/dagster_orchestration/hca_manage/billing_profile.py
+++ b/orchestration/dagster_orchestration/hca_manage/billing_profile.py
@@ -1,0 +1,48 @@
+import argparse
+import logging
+from typing import Optional
+import uuid
+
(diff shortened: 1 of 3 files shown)


eparkmonster (approved):
LGTM

quazi-broad (approved):
nice, lgtm. I'm wondering if any projects could take advantage of this code in the future as we see more users on Terra. I guess if that happens we can move tools like this out to dagster_utils so it can be used with other projects

aherbst-broad:
@quazi-broad yeah, exactly. I think this might fit in a "monster-tools' repo or some such thing.

PR #141, opened 2021-06-09
[DSPDC-1766] Thread through region when creating a dataset

aherbst-broad:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1766)
We should not hardcode "US" when creating datasets as we will not _always_ want them in that multi-region.

## This PR
* Parameterizes the region
* Updates affected tests


raaidbroad (approved):

quazi-broad (approved):
Looks good

PR #143, opened 2021-06-11
Fix Job tool to work in both the result and job use case

aherbst-broad:
## Why

(No relevant ticket)
I bumped into some wonkiness when polling jobs locally. We were expecting to be able to JSON serialize the result which in both "retrieveJobResult" and "retrieveJob" modes, which is not the case.

## This PR
* Only json serializes job results, not the job model
* Adds a .gitignore entry for autoenv which I've been using to good effect.
* Adds an "enumerate" mode for billing profiles

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.autoenv.zsh
 target/
 .scalafmt.conf
 .idea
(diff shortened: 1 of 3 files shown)


raaidbroad (approved):

quazi-broad (approved):
Looks good, will be useful to have that level of detail in the logs

PR #148, opened 2021-06-17
Do not pass certain params when calling jade prod

aherbst-broad:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1791)
We ended up provisioning the LungMap dataset in the Jade prod env. This uncovered some incompatibilities in our dataset creation code, specifically two parameters that are not supported in that environment.

## This PR
* Conditionally removes the parameters if the dataset creation call is being made against Jade `prod`

### DIFF
--- a/ops/helmfiles/dagster/forward_ports.sh
+++ b/ops/helmfiles/dagster/forward_ports.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 # This script forwards port 8080 to the Kubernetes cluster running our Dagster install so you can access
 # the Dagit dashboard from your machine.
+set -e
 ENV=${1:-dev}
 
 echo $ENV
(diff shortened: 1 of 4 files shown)


ops/helmfiles/dagster/forward_ports.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 # This script forwards port 8080 to the Kubernetes cluster running our Dagster install so you can access
 # the Dagit dashboard from your machine.
+set -e
```
aherbst-broad, line 4: This is unrelated to the dataset changes, but:

Without this set, the gcloud k8s connection setup a few lines down can fail due to invalid creds (i.e., if you're attempting to connect to dev using your prod creds, or vice-versa). This led me to attempting to run prod workflows in dev. This change will cause the script to fail.

raaidbroad (approved):
nice and clean

orchestration/dagster_orchestration/hca_manage/dataset.py
```diff
@@ -184,6 +187,12 @@ def create_dataset(
             "region": region,
             "cloudPlatform": "gcp"
         }
+
+        if env == 'prod':
```
raaidbroad, line 191: not important at all, but consider using double quotes here to be consistent with the rest of the strings in our format

eparkmonster (approved):

PR #155, opened 2021-06-28
Dual write argo + dagster jobs

aherbst-broad:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1834)
We would like to dual-write imports through both the dagster and argo pipelines and verify that we are getting the same results down both paths.

## This PR
* Adds a dual_write tool that will dispatch imports through both pipelines
* Adds a mode that will diff two HCA datasets
## Checklist


orchestration/dagster_orchestration/hca_manage/validation.py
```diff
@@ -19,17 +17,6 @@ def fetch_schema(path: str) -> Any:
     return raw_schema
 
 
-@dataclass
-class GsBucketWithPrefix:
```
aherbst-broad, line 23: I extract this to the `contrib.gcs` module for use elsewhere.

orchestration/dagster_orchestration/hca_orchestration/pipelines/load_hca.py
```diff
@@ -34,7 +35,7 @@
 dev_mode = ModeDefinition(
     name="dev",
     resource_defs={
-        "beam_runner": preconfigure_resource_for_mode(k8s_dataflow_beam_runner, "dev"),
+        "beam_runner": preconfigure_resource_for_mode(dataflow_beam_runner, "dev"),
```
aherbst-broad, line 38: This still runs the transformation jobs in dataflow, but I didn't want to muck with k8s locally so I've switched this to just dispatch the dataflow jobs via sbt.

quazi-broad (approved):
Changes look good and make sense to me.

Was the script already run against all the datasets in the ticket? Mock controlled access staging area, staging areas, and dcp6 staging areas?

aherbst-broad:
@quazi-broad yes, things looked good after those runs. I plan on testing a bit more but I think we are in a good spot for the next dcp release.

PR #156, opened 2021-07-01
Auto add monster as a steward when creating datasets

quazi-broad:
## Why
Ensure that _monster@firecloud.org_ is auto added as a policy member when creating datasets in order to make sure the whole team can access datasets, snapshots, and profiles in TDR.

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1839)

## This PR
- Added the monster team email as the default value to the the policy-members CLI arguments for creating a dataset
- Added an init method to add the monster team email to a list of stewards which is included when adding policy_members
- Updated the unit test to make sure we auto added the monster team email and called the add dataset policy member function 3 times

## Checklist

### DIFF
--- a/orchestration/dagster_orchestration/hca_manage/dataset.py
+++ b/orchestration/dagster_orchestration/hca_manage/dataset.py
@@ -4,7 +4,7 @@ import logging
 import os
 import subprocess
 import sys
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from typing import Optional, Any
 from re import search
 
@@ -33,7 +33,8 @@ def run(arguments: Optional[list[str]] = None) -> None:
     dataset_create.add_argument("-n", "--dataset_name", help="Name of dataset to create.", required=True)
     dataset_create.add_argument("-b", "--billing_profile_id", help="Billing profile ID", required=True)
     dataset_create.add_argument(
-        "-p", "--policy-members", help="CSV list of emails to grant steward access to this dataset"
+        "-p", "--policy-members", default='monster@firecloud.org',
+        help="CSV list of emails to grant steward access to this dataset (default: monster@firecloud.org)"
     )
     dataset_create.add_argument("-s", "--schema_path", help="Path to JSON schema", required=False)
     dataset_create.add_argument("-r", "--region", help="GCP region for the dataset", required=True)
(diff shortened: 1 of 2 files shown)


quazi-broad:
Not sure if I did this right, all feedback is appreciated.
I feel like adding the monster email to be the default for the CLI arg AND using that init method is redundant?
Was also wondering if there was a good way to add a unit test to do a check to look for the exact string of the monster email...

orchestration/dagster_orchestration/hca_manage/dataset.py
```diff
@@ -113,6 +114,13 @@ def _query_dataset(args: argparse.Namespace) -> None:
 class DatasetManager:
     environment: str
     data_repo_client: RepositoryApi
+    steward_list: list[str] = field(init=False)
+
+    def __post_init__(self) -> None:
+        self.steward_list = {
```
aherbst-broad, line 120: ⛏️ This is a dict, not a list. Perhaps name this `self.stewards`?

orchestration/dagster_orchestration/hca_manage/tests/test_dataset_manager.py
```diff
@@ -51,7 +51,7 @@ def test_create_dataset_with_policy(self):
         )
 
         self.manager.data_repo_client.create_dataset.assert_called_once()
-        self.assertEqual(self.manager.data_repo_client.add_dataset_policy_member.call_count, 2)
+        self.assertEqual(self.manager.data_repo_client.add_dataset_policy_member.call_count, 3)
```
aherbst-broad, line 54: We can make this assertion a bit stronger and be sure that `monster@firecloud.org` was indeed added:
```suggestion
        calls = [
            call('fake_dataset_id', policy_name='steward', policy_member={'email':'monster@firecloud.org'}),
            call('fake_dataset_id', policy_name='steward', policy_member={'email':'abc@example.com'}),
            call('fake_dataset_id', policy_name='steward', policy_member={'email':'def@example.com'}),
        ]
        self.manager.data_repo_client.add_dataset_policy_member.assert_has_calls(calls, any_order=True)
```

aherbst-broad (approved):
lgtm, the test could use a stronger assertion

eparkmonster (approved):

orchestration/dagster_orchestration/hca_manage/dataset.py
quazi-broad: Updated the name to be "stewards" and I think we are actually initializing it as a set, and later on the "add_policy_members" method takes in the policy_members parameter as a set[str] (thanks mypy)

orchestration/dagster_orchestration/hca_manage/tests/test_dataset_manager.py
quazi-broad: thank you! didn't realize we could check for the actual calls themselves using the mock

PR #157, opened 2021-07-06
[DSPDC-1827] Validate snapshot name

aherbst-broad:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSDPC-1827) There is a spec for HCA snapshot names, we should adhere to it when creating snapshots.

## This PR
* Adds a check against the regex from the DCP2 spec and raises if the supplied name does not match
* Updates dataset name validation code to be consistent with the approach for the snapshot name validation
* Adds and updates tests
## Checklist

### DIFF
--- a/orchestration/dagster_orchestration/hca_manage/snapshot.py
+++ b/orchestration/dagster_orchestration/hca_manage/snapshot.py
@@ -2,6 +2,7 @@ import argparse
 from dataclasses import dataclass, field
 from datetime import datetime, date
 import logging
+import re
 import sys
 from typing import Optional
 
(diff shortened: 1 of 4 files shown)


orchestration/dagster_orchestration/hca_manage/snapshot.py
```diff
@@ -131,7 +139,9 @@ def submit_snapshot_request_with_name(self, snapshot_name: str) -> JobId:
     def snapshot_name(self, qualifier: Optional[str] = None, snapshot_date: Optional[date] = None) -> str:
         snapshot_date = snapshot_date or datetime.today().date()
         date_stamp = str(snapshot_date).replace("-", "")
-        if qualifier:
+        if not qualifier:
```
aherbst-broad, line 142: This prevents us from appending `None` to the snapshot name

eparkmonster (approved):
LGTM

quazi-broad (approved):
looks good to me, having specific exceptions for the snapshot and dataset name errors are an improvement over the generic value errors

PR #160, opened 2021-07-16
Add snapshot retrieval and snapshot policy update operations

aherbst-broad:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1860)
We need the ability to add policy members to a snapshot.

## This PR
* Adds CLI options to our snapshot tool that enable querying and adding snapshot policy members
* Customizes the snapshot policy members based on a new `managed_access` bool and threads that through to the relevant pipeline code

## Checklist

### DIFF
--- a/orchestration/dagster_orchestration/hca_manage/snapshot.py
+++ b/orchestration/dagster_orchestration/hca_manage/snapshot.py
@@ -6,7 +6,7 @@ from re import search
 import sys
 from typing import Optional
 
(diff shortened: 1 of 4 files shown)


eparkmonster (approved):

quazi-broad (approved):
Looks good to me, I was wondering whether any of the add policy member functions could be abstracted between this and the dataset code but I see that we are using two completely different functions from jade's client for each of these.

PR #169, opened 2021-07-29
Add Slack Notifications for the load_hca Pipeline

quazi-broad:
## Why
We should get a notification when a load_hca import starts and finishes.
The notification should capture the staging area being imported, the target dataset, and the jade project.
[Relevant ticket](https://broadinstitute.atlassian.net/browse/dspdc-1872)

## This PR
Added slack hooks to the load_hca pipeline for kickoff, failure, and success states.
Added the slack and dagit_config to the list of resource_defs for all the pipeline modes.
Threaded the new slack hooks into the existing solids.

## Checklist


orchestration/dagster_orchestration/hca_orchestration/pipelines/load_hca.py
```diff
@@ -72,17 +83,96 @@
...
         "target_hca_dataset": target_hca_dataset,
-        "bigquery_service": mock_bigquery_service
+        "bigquery_service": mock_bigquery_service,
+        "slack": console_slack_client,
+        "dagit_config": preconfigure_resource_for_mode(dagit_config, "test")
+    }
+)
+
+
+def _base_slack_blocks(title: str, key_values: dict[str, str]) -> list[dict[str, object]]:
```
aherbst-broad, line 93: 🔧 This is a copypasta of this [code](https://github.com/DataBiosphere/hca-ingest/blob/09fcd224c8638a2a059f2ac8244c7ff7f438893e/orchestration/dagster_orchestration/hca_orchestration/pipelines/cut_snapshot.py#L90), what do you think about extracting a common method?

orchestration/dagster_orchestration/hca_orchestration/pipelines/load_hca.py
```diff
@@ -72,17 +83,96 @@
...
+        }
+    ]
+
+
+@success_hook(
+    required_resource_keys={'slack', 'target_hca_dataset', 'dagit_config'}
+)
+def import_start_notification(context: HookContext) -> None:
+    context.log.info(f"Solid output = {context.solid_output_values}")
+    job_id = context.solid_output_values["result"]
```
aherbst-broad, line 126: 🔧 We will not have a TDR job ID in this context (at the start of the pipeline). I'd say remove this and the corresponding dict entry below.

orchestration/dagster_orchestration/hca_orchestration/pipelines/load_hca.py
```diff
@@ -72,17 +83,96 @@
...
+    kvs = {
+        "Staging Area": context.solids.pre_process_metadata.input_prefix,
```
aherbst-broad, line 128: ❓ Interesting, I hadn't thought the input_prefix would be present in the context here. Have you verified this works?

orchestration/dagster_orchestration/hca_orchestration/pipelines/load_hca.py
```diff
@@ -72,17 +83,96 @@
...
     }
+
+    context.resources.slack.send_message(blocks=_base_slack_blocks("HCA Starting Import", key_values=kvs))
+
+
+@failure_hook(
+    required_resource_keys={'slack', 'target_hca_dataset', 'dagit_config'}
+)
+def import_failed_notification(context: HookContext) -> None:
+    if "result" in context.solid_output_values:
```
aherbst-broad, line 142: 🔧 Same comment about job_id in this context--we won't have it. Remove the references to job_id here as well.

orchestration/dagster_orchestration/hca_orchestration/pipelines/load_hca.py
```diff
@@ -72,17 +83,96 @@
...
 def load_hca() -> None:
-    staging_dataset = create_scratch_dataset(pre_process_metadata(clear_scratch_dir()))
-    result = import_data_files(staging_dataset).collect()
+    hooked_staging_dataset = create_scratch_dataset(pre_process_metadata(clear_scratch_dir())).with_hooks({import_start_notification})
+    hooked_result = import_data_files(hooked_staging_dataset).collect().with_hooks({import_failed_notification})
 
-    file_metadata_fanout(result, staging_dataset)
-    non_file_metadata_fanout(result, staging_dataset)
+    file_metadata_fanout(result, staging_dataset).with_hooks({import_failed_notification})
+    non_file_metadata_fanout(result, staging_dataset).with_hooks({import_failed_notification}).with_hooks({message_for_import_done})
```
aherbst-broad, line 178: 🤔 This is fine for now but it got me thinking:

This is exposing a bit of an awkward aspect of this DAG, namely that there is no "final" solid in the pipeline. We're banking on the non_file_metadata_fanout always being the final branch to complete but that might not always be the case. 

I think the real solution here will be to materialize an asset on completion and fire a sensor when such an asset materializes, but that's out of scope for this PR.

aherbst-broad (changes requested):
Looking good, I think remove the job_id references and we'll be set to ship.

orchestration/dagster_orchestration/hca_orchestration/pipelines/load_hca.py
quazi-broad: If I abstract this within the HCA project, should it stay within the hca_orchestration project/directory? My initial thought was to add it to the [dagster_utils slack resource](https://github.com/broadinstitute/dagster-utils/blob/main/dagster_utils/resources/slack.py) but I'm wondering if this would be usable across all projects pulling from dagster_utils or if this is specifically for HCA?

aherbst-broad: 🤔 I think let's leave it in HCA for now.

orchestration/dagster_orchestration/hca_orchestration/pipelines/load_hca.py
quazi-broad: I wasn't sure about where to pull the staging area from but I saw that we were importing that solid. Have not verified that this works because I was still getting tripped up on the fanout hooks

aherbst-broad (approved):

eparkmonster (approved):
LGTM

PR #170, opened 2021-08-02
Project copying checkpoint PR [DSPDC-1851]

aherbst-broad:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1851)
We need the ability to copy projects between HCA datasets. This is the first checkpoint PR of that effort.

## This PR
* Implements subgraph hydration, querying the links table for a given HCA project ID and determining all constituent entities that need to be copied over
* Implements data file ingestion, by pulling access info from Jade for each data file entry, synthesizing an ingestion control file and firing off the relevant Jade API request
* Implements basic tabular data ingestion
* Implements outdated version soft deletion by using a BQ [window function query](https://cloud.google.com/bigquery/docs/reference/standard-sql/analytic-function-concepts). This is a pattern I hope to replace our current HCA outdated ids code with as it's simpler (1 step) and semantically clearer.

## Checklist


orchestration/dagster_orchestration/hca_orchestration/contrib/bigquery.py
```diff
@@ -33,6 +33,28 @@ def build_query_job(
 
         return query_job
 
+    def build_query_job(
```
aherbst-broad, line 36: This is a helper that we (somewhat surprisingly) had not needed until now: fire off a simple query to BQ and return the job, no external destination or table overwrite.

orchestration/dagster_orchestration/hca_orchestration/pipelines/copy_project.py
```diff
@@ -1,32 +1,61 @@
...
+    name="dev",
+    resource_defs={
+        "bigquery_client": bigquery_client,
+        "data_repo_client": preconfigure_resource_for_mode(jade_data_repo_client, "dev"),
+        "gcs": google_storage_client,
+        "scratch_config": scratch_config,
+        "bigquery_service": bigquery_service,
+        "hca_project_copying_config": hca_project_copying_config,
+        "target_hca_dataset": target_hca_dataset,
+        "load_tag": load_tag
```
aherbst-broad, line 28: `load_tag` is reused from the base load HCA pipeline.

orchestration/dagster_orchestration/hca_orchestration/solids/copy_project/inject_file_ids.py
```diff
@@ -0,0 +1,14 @@
...
+    AbstractComputeExecutionContext,
+)
+
+
+@solid(
+    required_resource_keys={"gcs", "data_repo_client"},
+    input_defs=[InputDefinition("start", Nothing)]
+)
+def inject_file_ids(context: AbstractComputeExecutionContext) -> str:
+    gcs = context.resources.gcs
```
aherbst-broad, line 12: TODO

orchestration/dagster_orchestration/hca_orchestration/contrib/bigquery.py
quazi-broad: I understand that this new function allows us to perform a BQ query without having to build/autodetect an external config in the job_config and also doesn't write the results anywhere but I was a little confused about how this would work without a schema - is it just using the autodetect feature to do that?

quazi-broad (approved):
Overall changes look good to me, just had a specific question about a new function which I've added a comment about.

orchestration/dagster_orchestration/hca_orchestration/contrib/bigquery.py
aherbst-broad: Good question. In this context, there is a schema; we are just firing a vanilla query off against an existing, vanilla dataset. BigQuery has another mode that allows you to query files stored in GCS (as we do in `build_query_job_using_external_schema`). We are setting `autodetect=true` there to avoid having to package up and ship a schema for those GCS areas.

eparkmonster (approved):

PR #173, opened 2021-08-03
Fix CI slowness by setting request timeout on fixture dataset deletion

aherbst-broad:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1876)
CI is taking upwards of 30 minutes to complete. I've tracked this down to the fixture dataset deletion request we are making to TDR. Specifically, there is no timeout set in our calls to the client, so I believe `urllib3` under the hood is using the default system socket timeouts. It's still unclear why this call would take _minutes_.

## This PR
* Adds a 30 second timeout to the `delete_dataset` call. This causes a quick failure and then a retry which immediately succeeds.

## Checklist


quazi-broad (approved):
lgtm

eparkmonster (approved):

PR #177, opened 2021-08-12
Tool to find project row IDs in TDR

aherbst-broad:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1879)
There was an issue in DCP8 where a project was included for import that should not have been. The fix in these cases is to
find the related datarepo_row_ids in the `project` and `links` tables and submit them for soft deletion. This is an attempt
to script up the first part of that process so we don't have to adjourn to the BQ web console every time this happens.

## This PR
* Adds a `find_project_rows.py` script that outputs the `project` and `links` `datarepo_row_ids` in question.

## Checklist

### DIFF
--- a/ops/helmfiles/dagster/helmfile.lock
+++ b/ops/helmfiles/dagster/helmfile.lock
@@ -5,7 +5,7 @@ dependencies:
   version: 0.0.6
 - name: dagster
   repository: https://dagster-io.github.io/helm
-  version: 0.12.13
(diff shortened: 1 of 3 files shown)


eparkmonster (approved):

quazi-broad (approved):
lgtm, this will be very useful for any soft deletes we may have to do

PR #178, opened 2021-08-12
Refactor table loading code

aherbst-broad:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1875)
The dagster HCA import pipeline was ignoring updated versions of metadata entities. While implementing the bugfix, I've taken the opportunity to clean up our common table loading code and clean things up a bit. 

## This PR
* Implements the bugfix in `load_table.py` by changing the join point from `primary_key` to `primary_key + version`
* Adds an end-to-end test that exercises the update functionality
* Refactors `load_table` to be more obvious and use a new data_repo_service abstraction I've introduced (see comments in the PR)

## Checklist


orchestration/dagster_orchestration/hca_manage/dataset.py
```diff
@@ -150,7 +150,8 @@ class DatasetManager:
     def __post_init__(self) -> None:
         self.stewards = {
             "dev": {"monster-dev@dev.test.firecloud.org"},
-            "prod": {"monster@firecloud.org"}
+            "prod": {"monster@firecloud.org"},
+            "real_prod": {"monster@firecloud.org"},
```
aherbst-broad, line 154: Unrelated to the overall PR, this makes the `dataset.py` tool functional in the "real" prod environment.

orchestration/dagster_orchestration/hca_orchestration/contrib/bigquery.py
```diff
@@ -1,25 +1,24 @@
-from typing import Optional
```
aherbst-broad, line 1: Most (all?) of the changes in this service are to start returning `RowIterator` instances rather than `Job` instances. That is, rather than forcing the client to call `.result()` on the returned `Job` to actually do the thing (make the query do the extraction, whatever), the service will now do it and return the materialized results. 

This makes a cleaner interface and means less of a leaky abstraction (i.e., client code of this service should not have to know or care about dispatching jobs on BQ, and instead only have to care about the actual results).

orchestration/dagster_orchestration/hca_orchestration/contrib/data_repo.py
```diff
@@ -0,0 +1,58 @@
+"""
```
aherbst-broad, line 1: As the comment says, this is an attempt to abstract over the raw jade data repo client and provide some insulation and testability. Importantly, all operations here automatically poll on jobs rather than relying on the consuming code to handle that.

orchestration/dagster_orchestration/hca_orchestration/pipelines/load_hca.py
```diff
@@ -103,6 +109,11 @@ def import_start_notification(context: HookContext) -> None:
     context.resources.slack.send_message(blocks=key_value_slack_blocks("HCA Starting Import", key_values=kvs))
 
 
+@solid
+def terminal_solid(results1: list[Optional[JobId]], results2: list[Optional[JobId]]) -> None:
```
aherbst-broad, line 113: This solid is a placeholder that will let us ping slack on pipeline completion.

orchestration/dagster_orchestration/hca_orchestration/solids/load_hca/data_files/load_data_metadata_files.py
```diff
@@ -117,11 +117,9 @@ def ingest_metadata_for_file_type(
     return file_metadata_fanout_result
 
 
-@composite_solid
```
aherbst-broad, line 120: I've simplified things in this branch of the pipeline to eliminate the anti-pattern of a composite calling another composite.

orchestration/dagster_orchestration/hca_orchestration/solids/load_hca/load_table.py
```diff
@@ -1,56 +1,122 @@
 import logging
```
aherbst-broad, line 1: This is a fairly sizable refactor of the table loading code. The main points of interest:

1. Reduced the granularity of the solids to avoid the awkward `yield result -> check result to conditionally dispatch downstream solid` pattern. In retrospect that pattern has two related downsides: a) it's difficult to test b) it's difficult to debug operationally. Rolling all of that conditional logic into a single (beefier) python method makes this code less surprising.

2. Switches to consuming a `data_repo_service` which lets us get rid of all the job polling code that was sprinkled throughout.

orchestration/dagster_orchestration/hca_orchestration/solids/load_hca/load_table.py
```diff
@@ -1,56 +1,122 @@
...
+    logging.info(f"New rows for metadata type {metadata_type}, soft deleting outdated versions...")
+    maybe_outdated_job_id: Optional[JobId] = clear_outdated(
+        scratch_config,
+        target_hca_dataset,
+        metadata_type,
+        bigquery_service,
+        data_repo_service,
+        gcs_client
+    )
+    return maybe_outdated_job_id
```
aherbst-broad, line 84: I think a likely `todo:` here is to return the number of new rows instead of the job ID, but that would require further downstream refactoring to accept that result.

orchestration/dagster_orchestration/hca_orchestration/solids/load_hca/load_table.py
```diff
@@ -1,56 +1,122 @@
...
         joined_table_name: str,
         scratch_config: ScratchConfig,
-        target_hca_dataset: HcaDataset,
+        target_hca_dataset: TdrDataset,
         scratch_dataset_name: HcaScratchDatasetName,
         bigquery_service: BigQueryService
 
 ) -> None:
-    datarepo_key = f"{primary_key} as datarepo_{primary_key}"
+    datarepo_key = f"{primary_key} as datarepo_{primary_key}, version as datarepo_version"
```
aherbst-broad, line 98: This is the actual bugfix; we weren't joining on the version before so the code was interpreting the presence of _any_ version of an entity as a signal to skip any updates.

orchestration/dagster_orchestration/hca_orchestration/solids/load_hca/load_table.py
```diff
@@ -1,56 +1,122 @@
...
     fq_dataset_id = target_hca_dataset.fully_qualified_jade_dataset_name()
 
     query = f"""
-    SELECT J.datarepo_row_id, S.*, {datarepo_key}
-    FROM {metadata_type} S FULL JOIN `{target_hca_dataset.project_id}.{fq_dataset_id}.{metadata_type}` J
-    USING ({primary_key})
+    SELECT existing.datarepo_row_id, staged.*, {datarepo_key}
+    FROM {metadata_type} staged FULL JOIN `{target_hca_dataset.project_id}.{fq_dataset_id}.{metadata_type}` existing
+    USING ({primary_key}, version)
```
aherbst-broad, line 104: More bugfix, joining on version as well as the primary key.

orchestration/dagster_orchestration/hca_orchestration/solids/load_hca/load_table.py
```diff
@@ -211,178 +227,80 @@ def start_load(
...
         bigquery_service: BigQueryService
-) -> None:
+) -> str:
     fq_dataset_id = target_hca_dataset.fully_qualified_jade_dataset_name()
     jade_table = f"{target_hca_dataset.project_id}.{fq_dataset_id}.{table_name}"
-    destination = f"{scratch_dataset_name}.{outdated_ids_table_name}"
+    out_path = f"{scratch_config.scratch_area()}/outdated-ids/{table_name}"
 
     query = f"""
+    EXPORT DATA OPTIONS(
```
aherbst-broad, line 251: This is a shortcut to let us skip exporting data to a table in our scratch dataset and instead just put the results directly at the desired path.

orchestration/dagster_orchestration/hca_orchestration/solids/load_hca/load_table.py
```diff
@@ -211,178 +227,80 @@ def start_load(
...
-
-
-def _export_outdated(
-        metadata_type: MetadataType,
-        outdated_ids_table_name: str,
-        scratch_dataset_name: HcaScratchDatasetName,
-        scratch_config: ScratchConfig,
-        bigquery_service: BigQueryService
-) -> None:
-    out_path = f"{scratch_config.scratch_area()}/outdated-ids/{metadata_type}/*"
```
aherbst-broad, line 263: Removing this code is what the `EXPORT DATA OPTIONS` shortcut mentioned above gets us.

orchestration/dagster_orchestration/hca_orchestration/solids/load_hca/non_file_metadata/load_non_file_metadata.py
```diff
@@ -40,6 +40,9 @@ class NonFileMetadataTypes(Enum):
     {"metadata_types": NonFileMetadataTypes, "prefix": "metadata"})
 
 
-@composite_solid
-def non_file_metadata_fanout(result: list[JobId], scratch_dataset_name: HcaScratchDatasetName) -> Nothing:
-    return ingest_non_file_metadata_type(result, scratch_dataset_name).map(load_table)
+@composite_solid(
+    output_defs=[DynamicOutputDefinition(dagster_type=Optional[JobId])]
+)
+def non_file_metadata_fanout(result: list[JobId], scratch_dataset_name: HcaScratchDatasetName) -> Optional[JobId]:
```
aherbst-broad, line 46: Same refactoring as the `load_data_metadata_files.py` file, i.e,. removing unneeded composite -> composite structure.

orchestration/dagster_orchestration/hca_orchestration/tests/e2e/conftest.py
```diff
@@ -99,7 +99,7 @@ def load_hca_run_config(dataset_name: str, dataset_info: DatasetInfo):
             },
             "scratch_config": {
                 "config": {
-                    "scratch_bucket_name": "broad-dsp-monster-hca-dev-test-storage",
+                    "scratch_bucket_name": "broad-dsp-monster-hca-dev-temp-storage",
```
aherbst-broad, line 102: Our test data should always go in the temp storage bucket so it'll get auto deleted after 7 days. 🤦

orchestration/dagster_orchestration/hca_orchestration/tests/e2e/test_updated_staging_area.py
```diff
@@ -0,0 +1,73 @@
+from datetime import datetime
+import pytest
+from dagster import execute_pipeline
+from google.cloud.bigquery import QueryJobConfig, Client
+
+from hca_orchestration.pipelines import load_hca
+
+
+@pytest.mark.e2e
+def test_load_updated_staging_area(
```
aherbst-broad, line 10: This is the e2e updates test. It fires off an import of the same staging data as `test_load_hca.py`, then fires off an import of a staging area with an updated entity I created. We may want to consider rolling the assertions from `test_load_hca` into the first section of this test to avoid running the same slow-ish import twice, but I think this is ok for now.

orchestration/dagster_orchestration/hca_orchestration/tests/solids/load_hca/test_load_table.py
```diff
@@ -1,120 +1,100 @@
...
-    "test_load_table",
-    resource_defs={**test_mode.resource_defs}
-)
-load_table_test_mode.resource_defs["gcs"] = ResourceDefinition.hardcoded_resource(
-    FakeGCSClient(
-        buckets={test_bucket_name: test_bucket}
-    )
-)
-
-run_config = {
```
aherbst-broad, line 29: Rather than having bare hardcoded constants, I've wrapped all of these in `pytest.fixture` fixture functions below.

quazi-broad (approved):
changes look good. PR was well documented and comments and explanations for changes were easy to understand. thanks for the walkthrough

eparkmonster (approved):
LGTM

PR #181, opened 2021-08-20
Validation ingress pipeline HCA

eparkmonster:
## Why
Need to create validate ingress pipeline to utilize our validation tool for staging paths
[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1887)

## This PR
Creates the pipeline for the validation ingress
Has a failure hook to notify slack
## Checklist

### DIFF
--- a/orchestration/dagster_orchestration/hca_orchestration/pipelines/validate_ingress.py
+++ b/orchestration/dagster_orchestration/hca_orchestration/pipelines/validate_ingress.py
@@ -4,9 +4,9 @@ from dagster_utils.resources.google_storage import google_storage_client, mock_s
 from dagster_utils.resources.data_repo.jade_data_repo import jade_data_repo_client, noop_data_repo_client
 from dagster_utils.resources.slack import console_slack_client, live_slack_client
 
-from hca_manage.validation import run
 from hca_orchestration.config import preconfigure_resource_for_mode
 from hca_orchestration.contrib.slack import key_value_slack_blocks
+from hca_orchestration.solids.validate_ingress import pre_flight_validate, notify_slack_of_ingress_validation_results
 
 prod_mode = ModeDefinition(
     name="prod",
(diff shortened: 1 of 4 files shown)


orchestration/dagster_orchestration/hca_orchestration/pipelines/validate_ingress.py
```diff
@@ -0,0 +1,62 @@
...
+        "Dagit link": f'<{context.resources.dagit_config.run_url(context.run_id)}|View in Dagit>'
+    }
+    context.resources.slack.send_message(blocks=key_value_slack_blocks("Validation Ingress Failed", key_values=kvs))
+
+
+@pipeline(
+    mode_defs=[prod_mode, local_mode, test_mode, dev_mode]
+)
+def validate_ingress() -> None:
+    run()
```
quazi-broad, line 62: Good start! I think we might have to feed in the staging path to the run command.
`python hca_manage/validation.py -p gs://PATH` is how the validation tool is normally called manually.
I think we need to get the input_prefix or the path of the staging area from the context but I'm not sure of we can do that directly here in this pipeline or if we need to set up a validate_ingress solid first.

orchestration/dagster_orchestration/hca_orchestration/solids/validate_ingress.py
```diff
@@ -0,0 +1,46 @@
+from dagster import solid, String
+from dagster.core.execution.context.compute import AbstractComputeExecutionContext
+
+from hca_manage.common import ProblemCount
+
+from hca_manage.validation import run as validate
+
+@solid(
+    required_resource_keys={'data_repo_client'},
```
aherbst-broad, line 9: 🔧 This solid doesn't appear to need a data repo client, right? In which case I think we can remove the resource altogether from this pipeline.

aherbst-broad (changes requested):
There is a redundant jade data repo client that I think we can do without.

PR #182, opened 2021-08-20
load_hca completion slack notification

quazi-broad:
## Why
We should be notified in #monster-ci when a load_hca import starts and when it finishes.
[Relevant ticket](https://broadinstitute.atlassian.net/browse/dspdc-1872)

## This PR
Added a success hook to send a slack notification upon successful completion of the load_hca pipeline

## Checklist

### DIFF
--- a/orchestration/dagster_orchestration/hca_orchestration/pipelines/load_hca.py
+++ b/orchestration/dagster_orchestration/hca_orchestration/pipelines/load_hca.py
@@ -108,11 +108,26 @@ def import_start_notification(context: HookContext) -> None:
 
     context.resources.slack.send_message(blocks=key_value_slack_blocks("HCA Starting Import", key_values=kvs))
 
+@success_hook(
+    required_resource_keys={'slack', 'target_hca_dataset', 'dagit_config'}
+)
+def import_complete_notification(context: HookContext) -> None:
+    context.log.info(f"Solid output = {context.solid_output_values}")
+    kvs = {
+        # todo: how do I pull staging area out here (that is from the preprocess_metadata solid config)
+        # "Staging area": context.solid_config["input_prefix"],
+        "Target Dataset": context.resources.target_hca_dataset.dataset_name,
+        "Jade Project": context.resources.target_hca_dataset.project_id,
+        "Dagit link": f'<{context.resources.dagit_config.run_url(context.run_id)}|View in Dagit>'
+    }
+
+    context.resources.slack.send_message(blocks=key_value_slack_blocks("HCA Completed Import", key_values=kvs))
 
 @solid
 def terminal_solid(results1: list[Optional[JobId]], results2: list[Optional[JobId]]) -> None:
+    # todo: should there be a check for whether we get valid results?
+    # if (results1 is not None and results2 is not None):
     pass
-
 
 @pipeline(
     mode_defs=[prod_mode, dev_mode, local_mode, test_mode]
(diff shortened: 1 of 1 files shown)


quazi-broad:
Two open questions:

- How do I pull the staging area out for the import complete notification?

That item is pulled from the preprocess_metadata config by the import start notification function, which is able to access it because it is hooked into that particular solid. Is there a way or me to pass that staging area back or perhaps have terminal solid be able to see the preprocess_metadata config?

- Does the terminal_solid function have to have any logic that asserts that both result sets exist (so that both the file metadata and non file metadata fanouts were both successful)?

orchestration/dagster_orchestration/hca_orchestration/pipelines/load_hca.py
```diff
@@ -108,12 +108,27 @@ def import_start_notification(context: HookContext) -> None:
 
     context.resources.slack.send_message(blocks=key_value_slack_blocks("HCA Starting Import", key_values=kvs))
 
+@success_hook(
+    required_resource_keys={'slack', 'target_hca_dataset', 'dagit_config'}
+)
+def import_complete_notification(context: HookContext) -> None:
```
aherbst-broad, line 114: 🔧 I don't think this needs to be a hook but can rather go directly in the terminal solid code (or be called from there at least).

aherbst-broad (changes requested):

orchestration/dagster_orchestration/hca_orchestration/pipelines/load_hca.py
quazi-broad: Understood, this has been refactored to just send the slack message from the terminal solid

orchestration/dagster_orchestration/hca_orchestration/pipelines/load_hca.py
```diff
@@ -109,9 +109,20 @@ def import_start_notification(context: HookContext) -> None:
...
 
-@solid
-def terminal_solid(results1: list[Optional[JobId]], results2: list[Optional[JobId]]) -> None:
-    pass
+@solid(
+    required_resource_keys={'slack', 'target_hca_dataset', 'dagit_config'}
+)
+def terminal_solid(context: HookContext, results1: list[Optional[JobId]], results2: list[Optional[JobId]]) -> None:
+    # todo: do I need to log the solid output here?
+    # context.log.info(f"Solid output = {context.solid_output_values}")
```
quazi-broad, line 117: should I be logging anything here?

orchestration/dagster_orchestration/hca_orchestration/pipelines/load_hca.py
```diff
@@ -109,9 +109,20 @@ def import_start_notification(context: HookContext) -> None:
...
+    kvs = {
+        # todo: how do I pull staging area out here (that is from the preprocess_metadata solid config)
+        # "Staging area": context.solid_config["input_prefix"],
```
quazi-broad, line 120: I couldn't figure out how to obtain the staging_area needed for the load_hca completion notification. Maybe I could just store the staging_area name earlier in the script and reference that in the separate slack message blocks?

orchestration/dagster_orchestration/hca_orchestration/pipelines/load_hca.py
aherbst-broad: No, I'd remove this.

orchestration/dagster_orchestration/hca_orchestration/pipelines/load_hca.py
aherbst-broad: 🤔 The issue is that it's a part of a solid config rather than a globally available resource. I think the best approach would be to factor the input_prefix out of the related solid config and convert to a resource.

aherbst-broad (approved):

orchestration/dagster_orchestration/hca_orchestration/pipelines/load_hca.py
aherbst-broad: Actually, we can pull it from the context after all: `context.run_config["solids"]["pre_process_metadata"]["config"]["input_prefix"],`

PR #185, opened 2021-08-24
Fix CI breakage due to out of date poetry version

aherbst-broad:
## Why

CI is failing due to the python cryptography wheel being built from source, and consequently requiring a Rust installation. I believe this may be due to us using an outdated version of poetry.

## This PR
* Updates the version of poetry we use when building the docker user code image.
* Updates all other dependencies 

## Checklist


quazi-broad (approved):
lgtm

PR #193, opened 2021-08-27
Update release manifest tool to deep inspect metadata as well

aherbst-broad:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1892)
As part of the verification process prior to handing off a snapshot downstream for HCA, we should ensure that all metadata entities have been loaded and that the correct version is present.

## This PR
* Updates the `verify_release_manfiest.py` script to inspect the links + metadata directories in a given staging area, pull all entity_ids, versions and contents for each file and then verify that the corresponding data is present in the dataset.

## Checklist


quazi-broad (approved):
Deep inspect logic looks good.
How were we previously identifying issues with metadata entity loads/versions?

aherbst-broad:
@quazi-broad we were not :( The recently introduce updates E2E test should prevent issues with versions + updates, and the load_hca e2e test should provide coverage against our base load use case. This verification script gives us additional safety during the production data load process ("belts and suspenders")

PR #200, opened 2021-09-02
[DSPDC-1814] Add new mode and partitioning scheme for cut snapshot in dev

aherbst-broad:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1814)
We need to cut snapshots for each project ported to dev.

## This PR
* Adds a new mode to the cut snapshot pipeline that will compute the correct source dataset name and snapshot name given a source hca project ID, and an associated partition set over this mode.

## Checklist


quazi-broad (approved):
looks good, will start using this immediately for the dev refresh snapshots

PR #203, opened 2021-09-03
Setup queuedruncoordinator

aherbst-broad:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1894)
We should use a queuedruncoordinator instance so we can control the concurrency level for in flight jobs.

## This PR
* Sets up the queuedruncoordinator

## Checklist


quazi-broad (approved):
looks good, and we already confirmed this works properly in dev

PR #205, opened 2021-09-07
Sanity check that we have projects and links during dataset validation

aherbst-broad:
## Why

We should know if a dataset has 0 links or projects, as that indicates a bad or failed project copy.

## This PR
* Adds the needed checks to the dataset validation script

## Checklist


quazi-broad (approved):
looks good, this will be helpful in catching some of those errors sooner

PR #207, opened 2021-09-09
Add snapshot verification mode

aherbst-broad:
## Why

We want the ability to verify snapshots, not just datasets.

## This PR
* Removes the `datarepo_` prefix in relevant queries and conditionally formats the dataset name with it when not in snapshot verification mode
* Emits logging to stdout

## Checklist


quazi-broad (approved):
changes make sense to me

PR #212, opened 2021-09-16
Fix dev refresh copy project job

aherbst-broad:
## Why

We are specifying the `dev_refresh` mode for copy project backfills, but it does not exist for that pipeline leading to errors.

## This PR
* Do not specify mode
## Checklist


quazi-broad (approved):
makes sense!

PR #213, opened 2021-09-17
Prevent upgrade to 3.9.7

aherbst-broad:
## Why

There is a regression in python [3.9.7](https://bugs.python.org/issue45081). I believe a fix has been merged, but until it is released we cannot go past 3.9.6 due to the way our Beam Runner dataclasses subclass a base Protocol.

## This PR
* Adds a dependency specification for python to keep us off 3.9.7

## Checklist


quazi-broad (approved):
gotcha

PR #214, opened 2021-09-17
Extract common partition set loading mechanism

aherbst-broad:
## Why

Instead of committing dev refresh partition CSVs directly to source control, we should externalize them to GS (as we are doing for the DCP release mechanism)

## This PR
* Extracts a common "load partitions from a GS path" method and shifts the dev refresh code to use that.

## Checklist


quazi-broad (approved):
New gs_csv_partition_reader method looks good, pulling it out makes sense

PR #218, opened 2021-09-21
Add default snapshot stewards

aherbst-broad:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1910)
We should add add the monster team as a snapshot steward by default.

## This PR
* Adds a new solid and threads  in config + presets to add monster as stewards when using the cut_snapshot pipeline.

## Checklist


quazi-broad (approved):
lgtm

PR #220, opened 2021-09-21
Bump poetry version and bust dependency cache

aherbst-broad:
## Why

Our master CI is failing and it looks like it may be this [poetry bug](https://github.com/python-poetry/poetry/issues/4409). 

## This PR
* Bumps the poetry version and busts the dependency cache to ensure we reverify the hashes on all deps.

## Checklist


quazi-broad (approved):
lgtm

PR #221, opened 2021-09-27
Un-hardcode BQ location

aherbst-broad:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1917)
We have datasets in dev that are not always in the US multi-region. We should be able to import to dataset's in any region needed.

## This PR
* Un-hardcodes the `US` multi-region
* Determines the region of the target dataset at runtime and uses that to a) determine where to create the "staging" dataset and b) thread through the location to downstream import queries
* The staging bucket must _also_ be located in the same region. For now, I've created a staging bucket in "us-central1" that we can use. We'll need to determine if there's a better approach than a staging bucket per region.

## Checklist

### DIFF
--- a/orchestration/hca_orchestration/solids/load_hca/load_table.py
+++ b/orchestration/hca_orchestration/solids/load_hca/load_table.py
@@ -10,7 +10,7 @@ from google.cloud.storage import Client
 
 from hca_orchestration.contrib.bigquery import BigQueryService
 from hca_orchestration.contrib.data_repo.data_repo_service import DataRepoService
-from hca_orchestration.contrib.gcs import path_has_any_data
+from hca_orchestration.contrib.gcs import path_has_any_data, remove_empty_blobs
 from hca_orchestration.models.hca_dataset import TdrDataset
 from hca_orchestration.models.scratch import ScratchConfig
 from hca_orchestration.support.typing import HcaScratchDatasetName, MetadataType, MetadataTypeFanoutResult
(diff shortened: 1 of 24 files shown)


quazi-broad (approved):
Changes to thread the new arg for region look good to me. Passing the region through will also be helpful in the future.

orchestration/hca_orchestration/solids/load_hca/load_table.py
```diff
@@ -58,6 +58,8 @@ def load_table(
         logging.info(f"No data for metadata type {metadata_type}")
         return None
 
+    remove_empty_blobs(scratch_config.scratch_bucket_name, source_path, gcs_client)
```
aherbst-broad, line 61: Dataflow was staging empty blobs for file transfer requests, in addition to those with data. BQ was choking on the empty blobs when importing so we need to remove.

orchestration/hca_orchestration/solids/load_hca/utilities.py
```diff
@@ -1,20 +1,33 @@
...
-    required_resource_keys={'slack', 'target_hca_dataset', 'dagit_config'}
+    required_resource_keys={'slack', 'target_hca_dataset', 'dagit_config', 'data_repo_client'}
 )
 def terminal_solid(
         context: AbstractComputeExecutionContext,
         results1: list[Optional[JobId]],
         results2: list[Optional[JobId]]
 ) -> None:
+    target_hca_dataset: TdrDataset = context.resources.target_hca_dataset
+    check_result = CheckManager(
```
aherbst-broad, line 21: This is running validation automatically after every import (unrelated to the BQ region updates)

PR #222, opened 2021-09-27
Snapshot default public

aherbst-broad:
## Why

We should make snapshots public in DEV by default unless they are managed access.

## This PR
* Updates our load hca pipeline to use a real SAM instance
* Updates the test_load_hca configuration to fix the e2e tests in master

## Checklist


quazi-broad (approved):
changes lgtm

PR #225, opened 2021-09-28
Fire slack notification on validation failure

aherbst-broad:
## Why

We want to be notified in slack when validation fails on import.

## This PR
* Wires up the validation egress notification to the final step of the load_hca pipeline

## Checklist


orchestration/hca_manage/check.py
```diff
@@ -20,8 +20,7 @@ def run(arguments: Optional[list[str]] = None) -> None:
     parser.add_argument("-e", "--env", help="The Jade environment to target",
                         choices=["dev", "prod", "real_prod"], required=True)
 
-    parser.add_argument("-p", "--project", help="The Jade project to target")
-    parser.add_argument("-d", "--dataset", help="The Jade dataset to target", required=True)
+    parser.add_argument("-i", "--dataset-id", help="The Jade dataset ID to target", required=True)
```
aherbst-broad, line 23: This is unrelated; passing the project and dataset name is unnecessary. Since we'll always have the dataset ID, we can discover the name and google project from that by querying jade.

orchestration/hca_orchestration/contrib/dagster.py
```diff
@@ -57,3 +58,10 @@ def gs_csv_partition_reader(gs_partitions_bucket_name: str, pipeline_name: str,
             ))
 
     return partition_sets
+
+
+def short_run_id(run_id: Optional[str]) -> str:
```
aherbst-broad, line 63: This is being used in a bunch of spots, centralizing here.

orchestration/hca_orchestration/solids/load_hca/utilities.py
```diff
@@ -25,16 +26,22 @@ def validate_and_notify(
...
+        )
+        context.resources.slack.send_message(message)
+    else:
+        kvs = {
+            "Staging area": context.run_config["solids"]["pre_process_metadata"]["config"]["input_prefix"],
+            "Target Dataset": context.resources.target_hca_dataset.dataset_name,
+            "Jade Project": context.resources.target_hca_dataset.project_id,
+            "Dagit link": f'<{context.resources.dagit_config.run_url(context.run_id)}|View in Dagit>'
+        }
+        context.resources.slack.send_message(blocks=key_value_slack_blocks("HCA Completed Import", key_values=kvs))
```
quazi-broad, line 44: Is "HCA Completed Import" correct here? It was already this way before your changes but just wondering because I thought this step was just validating and notifying.

quazi-broad (approved):
Makes sense to use the jade client instead of passing in additional args. Pulling short_run_id out saved a ton of repeated code! The slack notification code looks good.

PR #228, opened 2021-09-29
Update release config

aherbst-broad:
## Why

The dcp release partition set was attempting to be parameterized by "mode" + partition; however dagster only passes a partition to the provided run config function, which won't work since we need to know which config file to load depending on the mode. So, given that the dcp releases are really only taking place in prod, I am punting for now and limiting the dcp release partition set to prod mode and hardcoding.

## This PR
* Hardcodes the mode for the dcp release partition set function to prod
* Adds some comments + renames a few methods


quazi-broad (approved):
LGTM! sorry, I thought I had already approved this in the morning

PR #231, opened 2021-10-05
Add runner SAs by default

aherbst-broad:
## Why

We should have the correct permissions and region on new datasets.

## This PR
* Adds the needed runner SAs + removes code that prevented explicit setting of region when creating a dataset in prod.


quazi-broad (approved):
👍🏽

PR #232, opened 2021-10-06
Validate input_prefix

aherbst-broad:
## Why

A bad input prefix will silently pass through dataflow and is difficult to debug downstream. We should fail the pipeline early when we encounter one.

## This PR
* Adds checks for `gs://` and no trailing slash, related tests.


quazi-broad (approved):
changes lgtm, it will be helpful to catch those invalid urls up front with this validation

PR #235, opened 2021-10-07
Remove dead code

aherbst-broad:
## Why

The egress validation pipeline is no longer used, we validate as part of the load_hca pipeline.

## This PR
* Removes the pipeline and associated tests


quazi-broad (approved):
Removal of the validate_egress code looks good to me.

PR #237, opened 2021-10-12
Fix dataset tool to make dataset policy members optional

aherbst-broad:
## Why

The policy members arg is optional for this tool.

## This PR
* Checks if the arg is present before attempting to parse it

## Checklist


quazi-broad (approved):
This resolved the issue I was seeing when creating a dataset using the CLI tool.
Will update the monster ops playbook to reflect the changes

PR #252, opened 2021-11-01
Add ability to add a snapshot policy member to all snapshots matching…

aherbst-broad:
## Why

We need to add SAs and other groups as readers to many snapshots occasionally, this automates that process.

## This PR
* Add an option to the `snapshot.py` script to bulk add a policy member to snapshots matching a filter string. 

## Checklist


quazi-broad (approved):
This will save time when we have to add policy members to snapshots

PR #257, opened 2021-11-02
Use in proc executor when cutting snapshots to avoid snapshot name ch…

aherbst-broad:
## Why

The snapshot name is dynamically determined when instantiating resources during pipeline execution. When the snapshot creation process crosses date boundaries, the snapshot name can change if we are using the multiprocess executor. 

## This PR
* Switches to the in-proc executor to avoid snapshot name changes

## Checklist


quazi-broad (approved):
Good catch

PR #259, opened 2021-11-16
Fix assertion in e2e test

aherbst-broad:
## Why

The staging area we are ingesting as part of this test only has analysis files, so the assertion is incorrect.
## This PR
* Fixes the assertion
## Checklist


quazi-broad (approved):
lgtm

PR #261, opened 2021-11-19
Update copy project pipeline to work with DCP1 projects

aherbst-broad:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1930)
Some projects in DCP1 have data file metadata that use differing file_ids but the same underlying file. The copy project pipeline was making the mistaken assumption that all data metadata files would be 1<->1 with data files. 

## This PR
* Updates the subgraph hydration portion of the pipeline to explicitly craft a target path for any data metadata file being ingested from the descriptor. We will ingest all data files required by the descriptor, which may lead to dupes.
* Attempts to save some work by adding check for the target path in the datarepo load history, before attempting to resolve the DRS access URL and queuing the file for ingestion

## Checklist

### DIFF
--- a/orchestration/hca_orchestration/solids/copy_project/subgraph_hydration.py
+++ b/orchestration/hca_orchestration/solids/copy_project/subgraph_hydration.py
@@ -19,13 +19,17 @@ from hca_orchestration.models.scratch import ScratchConfig
 from hca_orchestration.resources.hca_project_config import HcaProjectCopyingConfig
 from hca_orchestration.support.subgraphs import build_subgraph_nodes
 from hca_orchestration.support.typing import MetadataType
+import json
+
+from hca_orchestration.models.hca_dataset import TdrDataset
 
 
 @op(
     required_resource_keys={
         "bigquery_service",
         "hca_project_copying_config",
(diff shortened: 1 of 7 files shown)


orchestration/hca_orchestration/solids/copy_project/subgraph_hydration.py
```diff
@@ -70,66 +78,100 @@ def hydrate_subgraphs(context: AbstractComputeExecutionContext) -> set[DataFileE
...
-                                                       bigquery_service)
-
-    drs_objects = {}
-    entity_file_ids = {}
+    # construct a target path from each file entity row
+    target_path_to_drs_uri = {}
     for entity_type, rows in entity_rows.items():
         if not entity_type.endswith("_file"):
             continue
-        for row in rows:
```
aherbst-broad, line 86: This approach led to a reused file only being able to be linked to a single metadata file entry, the below approach explicitly queues up the file as many times as needed by creating a unique target path from the descriptor.

quazi-broad (approved):
Changes to allow a reused file to be linked as many times as needed look good.

PR #264, opened 2021-11-23
[DSPDC-1949] Direct data file ingestion

aherbst-broad:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1949)
We have very large projects in HCA prod that cause GCS rate limiting. Jade has set the permission on the source bucket for the
offending projects so we can perform a direct bulk file ingest out of TDR, rather than having to copy to a staging area first.

## This PR
* Adds a config schema value that will switch on direct copy by default. We still need to be able to turn it off in most cases as this permission will not be set everywhere
## Checklist

### DIFF
--- a/orchestration/hca_orchestration/tests/e2e/conftest.py
+++ b/orchestration/hca_orchestration/tests/e2e/conftest.py
@@ -57,7 +57,6 @@ def dataset_info(dataset_name, delete_dataset_on_exit, existing_dataset_id, data
     # setup, either create the dataset or re-use the existing one if passed in as a fixture
     if existing_dataset_id:
         logging.info(f"Existing dataset ID = {existing_dataset_id}")
-        logging.info("This dataset will not be deleted at the end of the test")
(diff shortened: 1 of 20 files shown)


orchestration/hca_orchestration/tests/e2e/conftest.py
```diff
@@ -114,6 +114,13 @@ def copy_project_config():
                     "region": "US"
                 }
             }
+        },
+        "solids": {
+            "direct_copy_from_tdr": {
```
quazi-broad, line 119: Should this be the name of the function containing this parameter (ingest_data_files)?

quazi-broad (approved):
LGTM - I noticed most of the changes were also in the other PR (263).

orchestration/hca_orchestration/tests/e2e/conftest.py
aherbst-broad: Yup, good catch.

PR #265, opened 2021-11-29
Add qualifier to target dataset (DCP1 projects)

aherbst-broad:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1949)
We want to have separate datasets for the dcp1 and dcp2 versions of the dcp1 projects (they are present in both datasets). 

## This PR
* Adds a qualifier parameter to the `target_hca_dataset` dataset resource that will optionally append the supplied qualifier to the target dataset name during resolution
* Adds a check for any failedFiles during data file ingestion (the final validation step was catching this as well,but catching earlier in the pipeline is always better)

## Checklist


quazi-broad (approved):
New logic to handle qualifiers looks good to me

PR #266, opened 2021-12-08
Pull in RFC3339 lib to work around hca-import-validation bug

aherbst-broad:
## Why
The hca-import-validation lib is not pulling in the correct RFC 3339 validation lib.  This means it is considering any field that is of format `date-time` as passing validation. In the case of this field:
```
"collection_time": {
      ....
      "oneOf": [
        {
          "format": "date-time"
        },
        {
          "pattern": "^((19|20)\\d\\d){1}([- \\./][01]\\d){0,1}([- \\./][0-3]\\d){0,1}$"
        }
      ]
    },
```
it is passing both checks and therefore failing the `oneOf` constraint.

## This PR
* Pulls in the needed validation lib until the hca-import-validation lib can be updated upstream.

## Checklist

### DIFF
--- a/orchestration/poetry.lock
+++ b/orchestration/poetry.lock
@@ -1609,6 +1609,17 @@ requests = ">=2.0.0"
 rsa = ["oauthlib[signedtoken] (>=3.0.0)"]
 
 [[package]]
+name = "rfc3339-validator"
+version = "0.1.4"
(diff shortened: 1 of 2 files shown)


quazi-broad (approved):
lgtm

PR #269, opened 2021-12-21
Chunk target paths check and add test

aherbst-broad:
## Why

There is a DCP1 project w/85K files; when we go to check if the files have already been loaded in the target new prod dataset we are blowing out the BQ API call size limit. 

## This PR
* Adjusts the existing file path check to chunk the calls and return the merged result
* Adds a test

## Checklist


quazi-broad (approved):
Chunking up the previously_loaded calls looks good to me.

PR #272, opened 2022-01-03
[DSPDC-1957] Chunk _fetch_file_entities call

aherbst-broad:
## Why

This query also requires chunking to avoid going over the BQ API request size limit.

## This PR
Chunks the call, adds a test.
## Checklist


quazi-broad (approved):
Batching logic looks good to me

PR #273, opened 2022-01-06
Fix BQ project ID for DCP2 migration backfills

aherbst-broad:
## Why

We are using the DCP12 snapshot as the source for DCP2 based projects, the BQ project needs to reflect this and not be the base dataset's BQ project ID.

## This PR
* Sets the correct BQ project ID
## Checklist


quazi-broad (approved):
looks good to me

PR #276, opened 2022-01-12
[DSPDC-1957] Do not clobber entity jsons that were exported in previous chunks

aherbst-broad:
## Why

The entity chunking is incorrectly overwriting previously exported data, leading to missing entities for large projects.

## This PR
* Assigns a unique name per chunk to avoid the clobbering behavior

## Checklist


quazi-broad (approved):
makes sense to me

PR #280, opened 2022-01-26
Chunk IDs for subgraph verification

aherbst-broad:
## Why

We are hitting the BQ request size limit for extremely large projects with many subgraphs.

## This PR
* Chunks up the offending query in the subgraph verification code.

## Checklist


quazi-broad (approved):
changes look good to me

PR #281, opened 2022-01-31
Add simple script to verify snapshots are copied

aherbst-broad:
Add simple script to check that all projects from a given source snapshot exist in real prod.

### DIFF
--- a/orchestration/hca_manage/common.py
+++ b/orchestration/hca_manage/common.py
@@ -1,17 +1,22 @@
 import argparse
 import csv
-from dataclasses import dataclass
-import logging
 import functools
+import logging
 import sys
+from dataclasses import dataclass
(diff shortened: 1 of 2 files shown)


orchestration/hca_manage/common.py
```diff
@@ -23,7 +28,8 @@
 
 sam_host = {
     "dev": "https://sam.dsde-dev.broadinstitute.org",
-    "prod": "https://sam.dsde-prod.broadinstitute.org"
+    "prod": "https://sam.dsde-prod.broadinstitute.org",
+    "real_prod": "https://sam.dsde-prod.broadinstitute.org"
```
aherbst-broad, line 32: Unrelated, this is needed to be able to manually set snapshots public in real prod via the snapshot CLI tool.

quazi-broad (approved):
Snapshot verification script logic looks good

PR #285, opened 2022-03-01
Full per-project load support

aherbst-broad:
## Why

We are importing to new prod so our release infrastructure needs to support the new loading methodology.

## This PR
* Adds a new `load_hca_per_project` job that knows how to determine a project_id and consequent dataset given a source staging path and
* Updates the manifest generator tool to understand the new `institution, project_id` manifest format. 

## Checklist


quazi-broad (approved):
Changes to accomodate the new loading workflow and process the institution and project_id format look good.

PR #287, opened 2022-05-23
Pin markupsafe version

aherbst-broad:
## Why

There is a breaking change in the latest version of markupsafe

## This PR
* Pins to the markupsafe to the last known good version
* Fixes the setup-gcloud action invocations as well


ncalvanese1 (approved):
Looks good to me!

PR #294, opened 2023-02-28
Di 20 monster hca dev env

bahill:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DI-20)
Adding Docker Compose Dev env to speed up and simplify dev env setup

## Checklist
- [ x] Documentation has been updated as needed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -115,10 +115,11 @@ RUN mv /hca_manage /orchestration/. \
 
 CMD ["bin/bash"]
 
+# to build and run locally
 # docker build -t us-east4-docker.pkg.dev/broad-dsp-monster-hca-dev/monster-dev-env/hca_ingest_compose_dev_env:<new_version> .
 # docker run --rm -it us-east4-docker.pkg.dev/broad-dsp-monster-hca-dev/monster-dev-env/hca_ingest_compose_dev_env:<new_version>
 
-# to push to Artifact Registry
+# to build and push to Artifact Registry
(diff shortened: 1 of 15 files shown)


bahill:
OK - time to fight with the linter

Dockerfile
```diff
@@ -0,0 +1,127 @@
+## Builds the hca-ingest dev env image
+
+# This is the Dagster user code deployment image
+# Dockerfile is located in orchestration/Dockerfile
+# FROM us.gcr.io/broad-dsp-gcr-public/monster-hca-dagster:latest
+FROM us.gcr.io/broad-dsp-gcr-public/monster-hca-dagster:WIP_bhill
```
aherbst-broad, line 6: Will this eventually point at non-WIP tag?

docker-compose.yaml
```diff
@@ -0,0 +1,11 @@
+services:
```
aherbst-broad, line 1: I'm curious why docker compose is needed if the only service being run is a single container.

aherbst-broad (commented):
I can't view the relevant ticket, can you provide a bit more context about why this is necessary vs running in a venv?

docker-compose.yaml
bahill: Using Docker compose is allowing me to work in my local git clone but use the cached dev env.
https://github.com/ArjanCodes/2022-docker

I could also have used --reload in a run command, but I like the syncing this method provides, and the configuration of the run via a file vs a long command line.

Dockerfile
bahill: Ok that's all set now. Thank you for catching that!

aherbst-broad (approved):
lgtm, thanks for the clarification.

PR #295, opened 2023-03-02
DI 207 part1 - upgrade to python3.9.12 

bahill:
## Why

[Relevant ticket](https://broadworkbench.atlassian.net/browse/DI-207)

## This PR
We need to update the project dependencies in order to accept the dependabot alerts - because we should and because Bernick requested that we do this.

Step 1
Github actions are failing because ubuntu-latest no longer supports Python 3.9.6
- updated to Python 3.9.12

## Checklist


.github/Dev/push-to-docker-bits.yaml
```diff
@@ -0,0 +1,33 @@
+name: Push to Docker bits
```
bahill, line 1: Wanted to iterated faster - this was just for testing and is in a dev folder now

aherbst-broad (approved):

PR #299, opened 2023-03-27
adding requirements.txt for sourceclear scans

bahill:
## Why

[SourceClear doesn't support poetry](https://broadinstitute.slack.com/archives/CADU7L0SZ/p1679948642433369?thread_ts=1679939402.459169&cid=CADU7L0SZ)

## This PR
Adding requirements.txt to /orchestration

## Checklist


.github/linters/.jscpd.json
bahill, line 1: This is a config file for the jscpd linter which helps catch copy paste errors in your code, but in this case I expect there to be matches in a requirements file, given the dependency trees, so I don't want the linter to look over the requirements.txt.

bahill:
FYI - https://broadworkbench.atlassian.net/browse/DI-240 - the requirements.txt will need to be re-generated when dependencies are updated, so an action will be created to do this when there is a PR to main in the hca-ingest repository

KutnerBroadie (approved):

KutnerBroadie (approved):

PR #307, opened 2023-08-14
DI-273 update scratch config to use new us multi region bucket

bahill:
## Why

[DI-273](https://broadworkbench.atlassian.net/browse/DI-273)

## This PR
The [diff run during the load portion of the HCA ingest pipeline](https://github.com/DataBiosphere/hca-ingest/blob/b882471b283af12ab44a77fadadf9291c6d56085/orchestration/hca_orchestration/solids/load_hca/data_files/load_data_files.py#L76) does an external join against the buckets and bq is balking because BigQuery & Dataflow were not touched during this move, leaving them in US multi-region.

The suggested solution is to create a new multi-region scratch bucket. This will contain only the scratch data we process during this interim period, thus still garnering the cost savings on the now single-region bucket, while unblocking this production workflow.

## Checklist

[DI-273]: https://broadworkbench.atlassian.net/browse/DI-273?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bahill:
Not sure why the IDE had old commit in here... I'll fix those when resolving the merge conflicts and makes sure to clear out my IDE

ncalvanese1 (approved):

aherbst-broad (approved):

PR #313, opened 2023-09-28
[trivy] Add Trivy action

zbedo:
This is still in progress


bahill (approved):
Looks good! and it seems to have run without finding anything to fix - so that's great!

For the record - I ignored the python-validation check failure as it had to do with a script added by another team member and is not related to this PR.
I also canceled the E2E tests as this does not touch the codebase.

ncalvanese1 (approved):

PR #320, opened 2023-10-20
FE-92 fix git action to use real poetry lock

bahill:
## Why

[Relevant ticket](https://broadworkbench.atlassian.net/browse/FE-92)

## This PR

- Updates the current generate requirements.txt git action so it no longer creates a poetry.lock file. Instead, it uses the lock file already committed to the repository and used in the user code image. This ensures that the requirements.txt is both up to date and matches what is in production.

- Fixed the linter git action so that it only runs on pushes to branches that aren't main

## Checklist


bahill:
Something is wrong with this PR - it just stopped reporting on the checks it is running - while they are running.

PR #324, opened 2024-01-18
FE-128 dedup GitHub actions

bahill:
## Why

[Relevant ticket FE-128](https://broadworkbench.atlassian.net/browse/FE-128)

## This PR
Added better Git Action logic to the main & dev build actions to allow them to co-exist in the workflows dir.

PR is needed to see if I've fixed the second issue of PRs not showing up in the conversation on PRs. - fixed had to fix some syntax & add the jobs to the branch protection rules.
https://stackoverflow.com/questions/69177488/how-to-set-github-actions-as-required-status-checks
80% fixed - the check status is not syncing with the latest versions of the jobs from updates to this PR. I will dig into that later - it's not critical.

## Checklist


bahill:
OK I still don't know why the checks above don't line up with the actions - but for now: 
https://github.com/DataBiosphere/hca-ingest/actions

aherbst-broad (approved):

.github/workflows/build_and_publish_main.yaml
```diff
@@ -1,10 +1,11 @@
-name: Main Validation and Release
+name: Main Validation and Release Workflow
 on:
-  push:
-    branches:
-      - main
+  pull_request_target:
+      types:
+        - closed
```
ahaessly, line 5: Technically this runs on any PR that is merged - so even a feature branch off of dev being merged to dev. Is this what is intended? If so, change the name of the workflow and file to remove "main". If you want it to run only for main then maybe merge with the validate pull request workflow.

.github/workflows/validate-pull-request.yaml
```diff
@@ -5,6 +5,7 @@ on:
       - main
 jobs:
   pr-validation:
+    name: PR Validation
```
ahaessly, line 8: maybe change the name of this workflow (and file) to indicate this only happens for main.

ahaessly (approved):
👍

.github/workflows/build_and_publish_main.yaml
bahill: added main

PR #326, opened 2024-01-29
FE-147 add sentry for production readiness interim dagster

bahill:
## Why

[Relevant ticket FE-147](https://broadworkbench.atlassian.net/browse/FE-147)

## Checklist


orchestration/Dockerfile
```diff
@@ -9,7 +9,8 @@ ENV PYTHONFAULTHANDLER=1 \
   PIP_NO_CACHE_DIR=off \
   PIP_DISABLE_PIP_VERSION_CHECK=on \
   PIP_DEFAULT_TIMEOUT=100 \
-  POETRY_VERSION=1.1.8
+  POETRY_VERSION=1.1.8 \
+  SENTRY_DSN=https://922d76ad2dceb6cd328436ae551ffb11@o54426.ingest.sentry.io/4506559533088768
```
aherbst-broad, line 13: Does this mean the pipeline is using the same DSN for dev and prod? Will that intermix errors between the two environments or is there a way to distinguish between the two the sentry setup?

orchestration/hca_orchestration/pipelines/cut_snapshot.py
```diff
@@ -36,6 +38,14 @@
 warnings.filterwarnings("ignore", category=ExperimentalWarning)
 
 
+SENTRY_DSN = os.getenv(
+    "SENTRY_DSN",
+    "",
+)
+if SENTRY_DSN:
+    sentry_sdk.init(dsn=SENTRY_DSN, traces_sample_rate=1.0)
```
aherbst-broad, line 46: The docs [state](https://docs.sentry.io/platforms/python/configuration/options/?original_referrer=https%3A%2F%2Fwww.google.com%2F#dsn) that the sdk will attempt to pull from the environment automatically and if it can't find the DSN it will just not send events. Given that, what do you think about simplifying this logic to a `sentry_sdk.init(traces_sample_rate=1.0)` and letting the SDK sort out the env var resolution?

aherbst-broad (commented):
Had a question about the SDK init. Can you fix up the description a bit too? It seems to be mis-formatted.

orchestration/Dockerfile
bahill: Yes - Sam has a setup for this [whereby each env gets a DSN](https://github.com/broadinstitute/terra-helmfile/pull/3502) which I like, but seemed like overkill for a throwaway. Plus, the only time dev runs is when I am testing the pipeline and hopefully don't have to do that again?
And - I'm still unsure of the utility here as it's all just duplicating the dagit errors which are also caught in the cloud console logs explorer.
What do you think?

orchestration/hca_orchestration/pipelines/cut_snapshot.py
bahill: That makes sense to me - I just copied this in from what [Zebrafish & SCP did.](https://github.com/search?q=org%3Abroadinstitute+sentry_sdk.init&type=code) - I don't have strong feelings either way.

bahill: Yeah Bond does it too
https://github.com/DataBiosphere/bond/blob/e53378b1563ef2275ff66b9e924c0a3f776ac711/main.py#L21

bahill: At this point... I'm inclined to leave it alone unless I need to touch this again. I'd have to rebuild, deploy to dev, test, then PR to main, which triggers the E2E tests, and then go through review and deploy to prod, then smoke test. That functionally takes me to the end of the week.
What do you think about the compromise of making a ticket for this (it's what I did with cleaning up the git actions)? I'll pick this back up if I have to update the codebase again?

aherbst-broad: Sure, ticket makes sense.

orchestration/Dockerfile
aherbst-broad: Yeah, I'm not thrilled with the idea of intermixing the environments but given the time constraints and desire to replatform probably not worth revisiting at this point.

aherbst-broad (approved):
🚢

kilonzi (approved):

orchestration/Dockerfile
bahill: FWIW - if it becomes an issue in this interim, I will address it and definitely will not intermix in the re-platform

orchestration/hca_orchestration/pipelines/validate_ingress.py
```diff
@@ -13,6 +16,13 @@
     pre_flight_validate,
 )
 
+SENTRY_DSN = os.getenv(
+    "SENTRY_DSN",
+    "",
+)
+if SENTRY_DSN:
+    sentry_sdk.init(dsn=SENTRY_DSN, traces_sample_rate=1.0)
```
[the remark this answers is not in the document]
bahill: 100% with you on that. I weighed it out - doing something like SCP does, but it only saved me like 4 lines of code, and I'm not changing this again, so I went with the brute force duct tape approach.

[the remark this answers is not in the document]
bahill: Also - re: Drew's point from above - it's basically the same number of lines actually fewer files if I were to leave out the `if SENTRY_DSN` bit and let Sentry just figure out its DSN from the env

zbedo:
@bahill can we setup SENTRY_DSN as an env var in another PR?

